PR #107, opened 2024-08-19
fix: print out the path where the image encode failed

liaoliaojun:
FIX: When I have an image in my dataset that is too small in pixels, the training breaks down, but I can't find the path to that image. For example, an image with this pixel value: 122 x 63


jaretburkett:
Thank you!

PR #109, opened 2024-08-19
Add `push_to_hub` to the trainer

apolinario:
Add the ability to push the trained models to Hugging Face directly (incl. the images in the gallery)


jaretburkett:
Thanks @apolinario ! This is great!

PR #155, opened 2024-09-05
Fix bug with modal script

advay-modal:
Hi, this fixes a bug with the `run_modal` script


advay-modal:
@jaretburkett

advay-modal:
Following up here @jaretburkett

PR #179, opened 2024-09-17
Update RunPod instructions

Randomblock1:
Current RunPod example has overkill storage, which costs more, and wheel wasn't installed in venv, making installation take much longer.


brandonbevans:
pip install wheel is absolutely essential. Thanks for this bro

Issue #190, opened 2024-10-04
Flux LORA: ValueError: text input must be of type `str` ... for `flux.encode_prompt`


ScreenGlass:
## Describe the bug

Thank uwu fow youw wib!

When trying to make a lora for flux, I keep  getting this error about prompting (full traceback at the end)

```
Error running job: text input must be of type `str` (single example), `List[str]` (batch or single pretokenized example) or `List[List[str]]` (batch of pretokenized examples).
```

The error comes from 

https://github.com/ostris/ai-toolkit/blob/28e6f00790e05cfd8acaaee190b436b74713709c/toolkit/stable_diffusion_model.py#L2005

putting a print for `prompt` before gives
```
flux prompt ["[trigger] holding a sign that says 'I LOVE PROMPTS!'"]
flux prompt [False]
```
The first prompt is normal, but the second is a boolean which I do not know from where it comes from

Full traceback, any help appreciated
```bash
$ python3 run.py config/foo.yml
Running 1 job
/home/ubuntu/ai-toolkit/venv/lib/python3.12/site-packages/albumentations/__init__.py:13: UserWarning: A new version of Albumentations is available: 1.4.17 (you have 1.4.15). Upgrade using: pip install -U albumentations. To disable automatic update checks, set the environment variable NO_ALBUMENTATIONS_UPDATE to 1.
  check_for_updates()
/home/ubuntu/ai-toolkit/venv/lib/python3.12/site-packages/controlnet_aux/mediapipe_face/mediapipe_face_common.py:7: UserWarning: The module 'mediapipe' is not installed. The package will have limited functionality. Please install it using the command: pip install 'mediapipe'
  warnings.warn(
/home/ubuntu/ai-toolkit/venv/lib/python3.12/site-packages/controlnet_aux/segment_anything/modeling/tiny_vit_sam.py:654: UserWarning: Overwriting tiny_vit_5m_224 in registry with controlnet_aux.segment_anything.modeling.tiny_vit_sam.tiny_vit_5m_224. This is because the name being registered conflicts with an existing name. Please check if this is not expected.
  return register_model(fn_wrapper)
[...]

PR #191, opened 2024-10-04
Fix default for `neg` in sample config

ScreenGlass:
fix https://github.com/ostris/ai-toolkit/issues/190


toolkit/config_modules.py
```diff
@@ -878,4 +878,4 @@ def log_image(self, image, count: int = 0, max_count=0):
         if self.logger is None:
             return
 
-        self.logger.log_image(image, count, self.prompt)
\ No newline at end of file
+        self.logger.log_image(image, count, self.prompt)
```
ScreenGlass, line 881: artefact from GitHub

PR #232, opened 2024-12-10
feat: add extension registry for dynamic extension loading

levi:
I wanted to be able to create custom extensions and load them without having to store them inside the cloned repo directory. This adds two entry points for this -- loading an extension class and loading a process which creates an anonymous extension class to reduce boilerplate.

This PR is more of a suggestion to get feedback. I can add examples, fix formatting, etc. if this is something you want to add to the toolkit.


levi:
Closing for #233

PR #233, opened 2024-12-10
feat: add extension registry for dynamic extension loading

levi:
I wanted to be able to create custom extensions and load them without having to store them inside the cloned repo directory. This adds two entry points for this -- loading an extension class and loading a process which creates an anonymous extension class to reduce boilerplate.

This PR is more of a suggestion to get feedback. I can add examples, fix formatting, etc. if this is something you want to add to the toolkit.

### DIFF
--- a/jobs/ExtensionJob.py
+++ b/jobs/ExtensionJob.py
@@ -1,15 +1,14 @@
 import os
 from collections import OrderedDict
 from jobs import BaseJob
-from toolkit.extension import get_all_extensions_process_dict
-from toolkit.paths import CONFIG_ROOT
+from toolkit.extension_registry import ExtensionRegistry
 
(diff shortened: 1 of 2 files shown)


levi:
Main feedback in using this so far is that it's a pain trying to extend any existing built in package without completely copying it to my working directory or modifying ai-toolkit to be an importable python package. The former defeats the initial use case for me (extend SDTrainer to have custom save hook behavior) and the latter is a significant change to the project.

@jaretburkett thoughts on making ai-toolkit importable? Something I've always wanted with the original kohya (my internal implementation is a package). ai-toolkit seems in a perfect spot for it to be both extensible like ComfyUI does with its local directories and as an extensible python package used in a larger training system.

PR #251, opened 2025-02-13
feat: now we can upload images during train into tb

vaskers5:


vaskers5:
Now we can see images from test prompts inside config in tensorboard

PR #263, opened 2025-03-04
Fix EMA config behaviour

majen-dev:
EMA could not be explicitly set to false, always defaulted to True, so long as ema_config was not None

new behaviour
if ema config exists and use_ema is not false result to True, otherwise False (behaviour before fix -> if ema_config is not None use_ema is always True, even when set false, otherwise False)


inflamously:
What I've also noticed is that when storing ema within a job, sometimes the config acquired two values like:
image

PR #274, opened 2025-04-01
`t csource`  --> `source` in linux install instructions

josephrocca:
I think this was a typo.


josephrocca:
Solved in another commit.

PR #278, opened 2025-04-16
Add Hidream support

jaretburkett:
Currently requires >37GB of vram. Working on getting that to 24, but no promises


RamonGuthrie:
Apologies if I'm not explaining myself correctly, but is 32GB Vram needed to try a Lora or train a whole model or is it both?

pcollins-engineer:
How much vram is needed to train hidream lora 3 months later? Still 36gb?

PR #283, opened 2025-04-25
Adding the ability in the UI to force stop jobs

fewjative:
Problem:
When problems arise during a training and it's not able to stop as expected, the database entry for the job is left as running and yet the job is not actually running. Further, when using the pause button, jobs in this broken state don't actually get their status changed. You will continue to see 'running' as the status but the info will change to 'Stopping job...'.

The real issue is due to this bug, you can't restart trainings that are already stopped and good to restart.

Repro:
Start training -> hard power off computer -> reboot computer and open UI

Solution:
Add a force stop button that modifies the status in the db to be appropriate. To be honest, it can be a little confusing initially to have both a stop and a pause button but I'm hoping the information prompted in the dialog helps avoid confusion.

image

image

### DIFF
--- a/ui/src/app/api/jobs/[jobID]/forcestop/route.ts
+++ b/ui/src/app/api/jobs/[jobID]/forcestop/route.ts
@@ -1,0 +1,19 @@
+import { NextRequest, NextResponse } from 'next/server';
+import { PrismaClient } from '@prisma/client';
+
+const prisma = new PrismaClient();
+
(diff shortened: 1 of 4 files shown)


fgordon-maker:
YES, we need this, especially for those running on runpod using the Ostris template. It's been countless times when stopping the job late in the training has resulted in it getting stuck in the "stopping" state in the DB and starting again is impossible.

Anothergazz:
This happens to me too. Being able to force status to stopped would save a lot of time

Issue #332, opened 2025-07-08
tensor size mismatch when setting batch_size>1 while training flux_kontext lora


yaoliliu:
## This is for bugs only

Did you already ask [in the discord](https://discord.gg/VXmU2f5WEU)?

No

You verified that this is a bug and not a feature request or question by asking [in the discord](https://discord.gg/VXmU2f5WEU)?

Yes

## Describe the bug

The code in diffusers embeddings:

`time_guidance_emb = timesteps_emb + guidance_emb`

When batch_size is 1, timesteps_emb's size would be [1, X] and guidance_emb's size would be [2, X]，which works fine with the pytorch tensor auto broadcasting.

However when batch_size is n(>1), auto broadcasting no longer takes effect and cause problems like this:

![problem](https://files.lsmcloud.top/blog6c7788566e109d70949b3dc353ae4e9b.png)

Maybe we need to ensure the guidance and timesteps the same size before entering diffusers' embedding layer code?

``` python
noise_pred = self.unet(
            hidden_states=latent_model_input_packed.to(
                self.device_torch, cast_dtype),
            timestep=timestep / 1000,
            encoder_hidden_states=text_embeddings.text_embeds.to(
                self.device_torch, cast_dtype),
            pooled_projections=text_embeddings.pooled_embeds.to(
                self.device_torch, cast_dtype),
            txt_ids=txt_ids,
            img_ids=img_ids,
            guidance=guidance,
            return_dict=False,
            **kwargs,
        )[0]
```

Eric7758:
I have also encountered the same issue. Currently, I am training with batch size=1

rotatorotator:
Does your code snippet fix the problem? I would like to train on bigger batch sizes asap since I have many image pairs 👍

yaoliliu:
> Does your code snippet fix the problem? I would like to train on bigger batch sizes asap since I have many image pairs 👍

No it doesn't. It's just where the problem starts. With the guidance being [2*N, X] and later defined timestep_embed [N, X]，the size mismatch occurs, I guess adding `timestep_embed.repleate_interleave(guidance.shape[0]/timestep_embed[0], dim=0)` might fix the issue, but I'm not sure. Guess we need someone more professional to fix this.

StartHua:
同样遇到

Caselles:
Having the same problem here.

hack-mans:
[...]

PR #343, opened 2025-07-18
fix: Guidance incorrect shape

davertor:
When training a Flux kontext lora with a batch_size>1, an error is produced due to the size mismatch between guidance and timestep shapes as reported in #332.

image

The solution is to expand the guidance tensor to the batch_size dimension, instead of the latents_model_input.shape[0]


jaretburkett:
Thank you

PR #431, opened 2025-09-21
Add Renaming and Captioning tools to Datasets

michaeltrefry:
I added the ability to rename files after they've been uploaded. By default it's datasetName-### (incrementing index), but there's several patterns you can create.

The big addition however is integrated JoyCaption. This allows you to use JoyCaption directly in the dataset page to generate captions. There are a whole bunch of options for style and naming characters as well as directing the model how it should describe.

To use JoyCaption you must start the service on your selected GPU. The first time it starts it will download the LLM to your huggingface cache. After that it should start pretty quickly. There's an option to stop the service when you're done captioning to save resources.

I hope this fits well into your system, it's something I just really thought would be useful so I built it (with a lot of help from Claude!)


JoeGaffney:
This seems like a good addition

PR #464, opened 2025-10-17
Fix get_prompt_embeds method for Qwen-image-edit

Rupam-M:
- Convert control image tensors to PIL Images
- Handle BFloat16 → Float32 conversion
- Permute tensor shape from (ch, h, w) → (h, w, ch)
- Scale tensor values [0,1] → [0,255] and convert to uint8
- Support single images and batches
- Cast width/height to int to prevent float/int issues

Fixes KeyError: 'pixel_values' and TypeError: unsupported ScalarType BFloat16

### DIFF
--- a/extensions_built_in/diffusion_models/qwen_image/qwen_image_edit.py
+++ b/extensions_built_in/diffusion_models/qwen_image/qwen_image_edit.py
@@ -176,6 +176,8 @@ class QwenImageEditModel(QwenImageModel):
         if self.pipeline.text_encoder.device != self.device_torch:
             self.pipeline.text_encoder.to(self.device_torch)
 
+        # Convert control images to PIL format if provided
+        pil_images = None
         if control_images is not None:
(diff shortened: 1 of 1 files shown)


Rupam-M:
Changes Made:

1. Fixed get_prompt_embeds method (lines 188-224)

Original issue: The method was passing pre-processed tensor images directly to self.pipeline.encode_prompt(), but the pipeline expects PIL Images.

Changes made:
Added conversion of control image tensors to PIL Images
Added .float() conversion to handle BFloat16 → Float32 (NumPy doesn't support BFloat16)
Convert tensor format from (ch, h, w) to (h, w, ch) using .permute(1, 2, 0)
Scale values from [0, 1] to [0, 255] and convert to uint8
Handle both single images and batches correctly
Cast width and height to int() to avoid float/int type issues

2.  (TypeError: Got unsupported ScalarType BFloat16) - 
Fixed by adding .float() conversion

PR #492, opened 2025-10-31
[README Update] Add setting to change huggingface model cache folder.

vreff:
Added:

This UI downloads large model files in order to train on your datasets. If you wish to change the directory at which this UI downloads base models like Qwen, Flux, and Wan, simply set `HF_HOME=/path/to/hf/cache` in your environment before building and running the UI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,8 @@ npm run build_and_start
 
 You can now access the UI at `http://localhost:8675` or `http://<your-ip>:8675` if you are running it on a server.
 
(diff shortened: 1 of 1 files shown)


ali-m-github:
Hi, where can I find the download path for models? I already have the flux dev model and don't want to redownload but can't find any folder to put it in.

vreff:
> Hi, where can I find the download path for models? I already have the flux dev model and don't want to redownload but can't find any folder to put it in.

~/.cache/huggingface on mac, not sure about other systems.

PR #521, opened 2025-11-24
Suppress spurious warning about osx-temperature-sensor

duggelz:
Running `npm run build_and_start` produces an error message
`Module not found: Can't resolve 'osx-temperature-sensor'`
on non-MacOS platforms.  This error is harmless but a nuisance.


CoraAI:
Same here,  seems related to system info or something, any help would be appreciated

duggelz:
Abandoning pull request due to inactivity.

PR #567, opened 2025-12-04
Add support for DGX OS

raziel2001au:
Adds support for devices running DGX OS (including DGX Spark). I **had to remove the _dw_pose_ dependency**, as there is no version that works with DGX OS (likely due to the lack of an ARM64 build of onnxruntime). Looking at the code, it seems like it will simply error out if the dependency is not found and prompt you to install it manually, from my discussion with Ostris it sounded like he'd be okay with this approach.

From toolkit/control_generator.py:
```python
       elif control_type == 'pose':
            self.debug_print("Generating pose control")
            if self.control_pose_model is None:
                try:
                    import onnxruntime
                    onnxruntime.set_default_logger_severity(3)
                except ImportError:
                    raise ImportError(
                        "onnxruntime is not installed. Please install it with pip install onnxruntime or onnxruntime-gpu")
                try:
                    from easy_dwpose import DWposeDetector
                    self.control_pose_model = DWposeDetector(
                        device=str(device))
                except ImportError:
                    raise ImportError(
                        "easy-dwpose is not installed. Please install it with pip install easy-dwpose")
```

There are also some issues where DGX OS can't identify which versions of various dependencies it needs, so I've created a separate dgx_requirements.txt to include the exact versions of those dependencies, it then includes the normal requirements.txt to install the rest of the dependencies. They could be merged into the existing requirements.txt, but I thought it would be cleaner to handle it this way, and isolate anything that is DGX OS specific to its own requirements file.

The instructions I've added advise to install the latest PyTorch version, which officially supports the DGX Spark, as the older version recommended on the main setup page is quite old, and is about three times slower than the current release of PyTorch, which likely has optimisations for Grace Blackwell systems.

Finally, I've added a separate readme with installation instructions for DGX OS, as it is a little more involved than it is on other systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,9 @@ pip3 install --no-cache-dir torch==2.7.0 torchvision==0.22.0 torchaudio==2.7.0 -
 pip3 install -r requirements.txt
 ```
 
+For devices running **DGX OS** (including DGX Spark), follow [these](dgx_instructions.md) instructions.
+
+
 Windows:
 
 If you are having issues with Windows. I recommend using the easy install script at [https://github.com/Tavris1/AI-Toolkit-Easy-Install](https://github.com/Tavris1/AI-Toolkit-Easy-Install)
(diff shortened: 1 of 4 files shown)


jaretburkett:
Thank you

PR #608, opened 2025-12-25
feat: Add sequential_load option for low-RAM environments

mandeep511:
This feature enables training Flux models on systems with limited RAM
(like Google Colab Free Tier) by loading text encoders first, caching
embeddings to disk, unloading text encoders, then loading the transformer.
This prevents having both loaded simultaneously, reducing peak RAM usage.

Changes:
- Add sequential_load config option to ModelConfig
- Defer transformer loading in stable_diffusion_model.py when enabled
- Add load_deferred_transformer() method to StableDiffusion class
- Integrate with SDTrainer to load transformer after TE caching/unload
- Add UI checkbox in SimpleJob.tsx for Flux models
- Add example config: train_lora_flux_low_ram.yaml

Usage:
  model:
    sequential_load: true
  train:
    cache_text_embeddings: true

### DIFF
--- a/config/examples/train_lora_flux_low_ram.yaml
+++ b/config/examples/train_lora_flux_low_ram.yaml
@@ -1,0 +1,132 @@
+# FLUX LoRA Training Config for Low-RAM Environments (e.g., Google Colab Free Tier)
+# 
+# This config enables training FLUX.1-dev on low-RAM systems by using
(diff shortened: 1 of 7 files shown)


james-s-tayler:
I feel like I'm hitting up against this exact need for this with LTX-2 training also. If you have "Unload TE" enabled, then in theory, this is the logical way to do things.

Issue #536, opened 2025-11-28
Flux 2 Image Edit LoRA training.


Astroburner:
[ostris_setup_file.txt](https://github.com/user-attachments/files/23828731/ostris_setup_file.txt)

## This is for bugs only

Did you already ask [in the discord](https://discord.gg/VXmU2f5WEU)?

Yes

You verified that this is a bug and not a feature request or question by asking [in the discord](https://discord.gg/VXmU2f5WEU)?

Yes

## Describe the bug

Runpod RTX6000 used with Ostris Toolkit
Model: Flux2 
Datasets:  Control 213 Images / Target 213 Images (with prompts)
attached you can find the config file. 

Question: after 2000 Steps there is no change in the sample images
Did I do something wrong with the settings?
 
Is there a simple quide for Flux2? I have experience with Ostris Toolkit in Quen Image Edit and Flux kontext Lora training. All worked well.

Astroburner:
PS: Goal was change food pictures (CGI, 3d, 2d, plastic and so on ....) to a realistic food

Astroburner:
Update: after 3500 Steps noch change in the sampel images, stoped Runpod

VandersonQk:
It looks like the training just isn't working. I tried training two LORAs, and both failed. Same as @Astroburner , I used image pairs (control and target with captions and the same resolution).

The first one (with 260 images and 5,000 iterations) was to transform 3D scenes into a realistic style, but it did not work at all. Mine had the opposite effect. Instead of learning the style, it "forgot" what the base model knew. It's just outputting the original image untouched.

The second Lora (600 images, 2,250 iterations) was supposed to remove text overlays from anime and manga, but it was either outputting abstract art (not noise) or doing random things, like darkening the image or removing random objects. In some cases, it was outputting the original image untouched.

The datasets are completely fine. I previously trained the same model for Kontext, and it worked perfectly.

My configuration file (runpod): [reaslistic_ai_toolkit_setting.txt](https://github.com/user-attachments/files/23835112/reaslistic_ai_toolkit_setting.txt)

Astroburner:
> The datasets are completely fine. I previously trained the same model for Kontext, and it worked perfectly.

Did the same thing with Qwen Image Edit, wortking perfectly. Lost already over 30 Bucks to runpod and nothing worked for flux2

jaretburkett:
[...]

PR #629, opened 2026-01-04
Fix Flux2 training ignoring single control images

relaxis:
## Problem

When training a Flux2 edit LoRA with a single control path (e.g., `control_path_1`), the control images are silently ignored during training. This causes the model to output random hallucinations instead of learning the expected input→output transformation.

### Root cause

In `flux2_model.py`, `get_noise_prediction()` only checks `batch.control_tensor_list`:

```python
if batch.control_tensor_list is not None:
    # process controls...
```

However, when using a single control path, the dataloader stores the control image in `batch.control_tensor` (singular), not `batch.control_tensor_list`. This is consistent with the behavior in `dataloader_mixins.py` lines 947-955:

```python
if len(control_tensors) == 1:
    self.control_tensor = control_tensors[0]  # single control goes here
elif self.use_raw_control_images:
    self.control_tensor_list = control_tensors  # only for multiple controls
```

## Solution

Add a fallback that normalizes `control_tensor` into the same list-of-lists format as `control_tensor_list` before processing. This matches the pattern already used in `qwen_image_edit_plus.py` (lines 241-245).

The fix handles:
- 4D tensors `(B,C,H,W)` - single control per batch item
- 5D tensors `(B,N,C,H,W)` - multiple controls per batch item  
- List inputs

## Testing

Tested with a deblur LoRA (200 blurry→sharp image pairs):
- **Before fix**: Model outputs random hallucinations, loss erratic (0.1-2.8)
- **After fix**: Model properly processes control images, loss stable (0.2-0.7), samples show correct transformation

## Related

- Similar to the control handling standardization in #461
- May explain some reports of Flux2 edit training "not working" like #536

### DIFF
--- a/extensions_built_in/diffusion_models/flux2/flux2_model.py
+++ b/extensions_built_in/diffusion_models/flux2/flux2_model.py
@@ -312,7 +312,30 @@ class Flux2Model(BaseModel):
             img_cond_seq_ids: torch.Tensor | None = None
 
             # handle control images
-            if batch.control_tensor_list is not None:
+            # Normalize control_tensor and control_tensor_list to a consistent format
+            control_items = getattr(batch, "control_tensor_list", None)
+
+            # Fallback: handle single control_tensor (used when only one control path is specified)
+            if control_items is None:
+                ct = getattr(batch, "control_tensor", None)
+                if ct is not None:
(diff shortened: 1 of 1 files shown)


jaretburkett:
This should be solved in the latest rework of the flux models.

PR #658, opened 2026-01-20
fix flux2 image editing training device mismatch failure

DefinitlyEvil:
Ensure img_cond_seq and img_cond_seq_ids are on the same device before concatenation.
(code is tested to be running correctly)

Added code: 
```py
                if img_input.device != img_cond_seq.device:
                  img_cond_seq = img_cond_seq.to(img_input.device)
                  img_cond_seq_ids = img_cond_seq_ids.to(img_input_ids.device)
```


jaretburkett:
This has been solved with another PR. Thank you.

PR #673, opened 2026-01-27
Fix Z-Image text embedding concatenation

Armychimp:
## Problem

When using Z-Image models with `cache_text_embeddings: true`, training fails with:

```
ValueError: Batch size of latents must be the same or half the batch size of text embeddings
```

## Root Cause

Z-Image uses 2D tensors `(seq_len, embed_dim)` where batch size = list length. The `concat_prompt_embeds()` function was concatenating these along dim=0, producing one large tensor instead of keeping them as separate list elements.

## Fix

Detect 2D tensor format in `concat_prompt_embeds()` and collect tensors into a flat list instead of concatenating.

## Testing

Tested with `Tongyi-MAI/Z-Image` with text embedding caching enabled.


pixel-magician:
The training speed has slowed down by ten times.

jaretburkett:
This should be resolves with another commit. Thank you

PR #682, opened 2026-01-29
Fix Z-Image text embedding concatenation

hexvalid:
Fix `concat_prompt_embeds()` to handle Z-Image's 2D tensor format (seq_len, embed_dim) where batch size = list length. Previously it concatenated tensors along dim=0, producing one large tensor instead of keeping them as separate list elements.

Without this fix, training with `cache_text_embeddings: true` fails with: "Batch size of latents must be the same or half the batch size of text embeddings"


pixel-magician:
After applying these codes, the training no longer reported errors, but the training speed decreased by nearly ten times.

PR #685, opened 2026-01-29
Fix UI dataset list to recognize symbolic links

diodiogod:
## Description
This PR fixes the dataset scanner in the UI to recognize symbolic links pointing to directories.

## Problem
The current implementation only checks for regular directories using `isDirectory()`, which doesn't follow symbolic links. This prevents users from organizing datasets using symlinks without duplicating data.

## Solution
- Check if a dirent is a symbolic link
- Use `fs.statSync()` to verify the symlink points to a directory  
- Gracefully handle broken symlinks with try/catch

## Changes
- Modified `ui/src/app/api/datasets/list/route.ts` to detect both regular directories and symbolic links

## Testing
Tested with symbolic links on Windows and they now appear correctly in the dataset list.

### DIFF
--- a/ui/src/app/api/datasets/list/route.ts
+++ b/ui/src/app/api/datasets/list/route.ts
@@ -11,10 +11,22 @@ export async function GET() {
       fs.mkdirSync(datasetsPath);
     }
 
-    // find all the folders in the datasets folder
+    // find all the folders in the datasets folder (including symlinks)
     let folders = fs
       .readdirSync(datasetsPath, { withFileTypes: true })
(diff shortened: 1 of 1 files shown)


Auryg:
I’m not the author of this PR, but I wanted to add that I’ve been using this change locally for several months on Windows and it has worked reliably for me through every update.

The change is small and wildly helpful for some workflows.  I don't have all of my datasets in one place for a few different reasons.

If helpful, I can open a fresh PR rebased on current main, though it's working well for me as-is.

PR #695, opened 2026-02-02
Fix onChange path for EMA Decay input

willhsmit:
Changes to the EMA Decay input don't get preserved when switching back and forth between Advanced and Simple view. I believe the onChange is not writing it correctly here; it updates correctly in the UI with this change.


jaretburkett:
Thank you

PR #699, opened 2026-02-05
AdamWFP8 support

saluzafa:
# Description
This PR adds support for AdamWFP8, the goal was to benchmark this optimizer for Z-Image Base LoRA trainings.
The optimizer Python script come from https://github.com/None9527/None_Z-image-Turbo_trainer/blob/omni/src/zimage_trainer/optimizers/adamw_fp8.py 🙂


troycjj:
The loss is still around 0.4, and nothing has improved.

PR #703, opened 2026-02-06
Change epsilon value to 1e-8 for optimizers & adamwfp8 and adamwbf16 …

NBSTpeterhill:
…optimizer addition

Updated epsilon value for various optimizers to improve numerical stability.


jaretburkett:
The epsilon value is set at 1e-6 instead of 1e-8 for BF16, which most modern models use during training. 

BF16 only has ~3 decimal digits of precision (8 exponent bits, 7 mantissa bits), so 1e-8 effectively rounds to zero in BF16 range making 1e-6 needed for numerical stability and preventing dividing by zero.

PR #707, opened 2026-02-10
Fixing resizing bug

abionda-sc:
In Qwen Image Edit and Flux 2, width and height are inverted during control image resizing.
This causes the LoRA to train with wrong images and create distorted results.
This PR solves the bug with a small code fix.


avan06:
I have also confirmed that the width and height calculations in the original get_prompt_embeds are indeed inverted.

jaretburkett:
Wow, amazing catch! Thank you.

PR #709, opened 2026-02-11
Add logging support for Comet

sebastianberns:
Runs can now be monitored with [Comet](https://www.comet.com). 

Add comet logging to your config:
```yaml
    logging:
      use_comet: true
      project_name: "example-project"
      run_name: "test-run"
```


sebastianberns:
Note: the default code logging is disabled, because it is not necessary to upload the toolkit run scripts.
However, the config is logged as hyperparameters.

PR #726, opened 2026-02-25
Fix FLUX2 Klein load-time VRAM spikes on low-memory GPUs.

mhofer1976:
Keep the transformer and Qwen text encoder off CUDA during initial load/quantization in low-VRAM mode so model startup avoids full-model OOM before offloading and quantization can take effect.

## Summary
- avoid moving the full FLUX2 transformer to CUDA before quantization, which caused startup OOM on low-memory GPUs
- keep the transformer/text-encoder on CPU during low-VRAM model preparation and only move as needed
- use `qtype_te` for Klein Qwen text-encoder quantization instead of the transformer `qtype`

## Test plan
- [x] Reproduce prior failure on FLUX2 Klein 9B at `Loading transformer` with low VRAM settings
- [x] Verify loader no longer calls full-model CUDA move before quantization in the FLUX2 path
- [x] Verify Klein TE path no longer eagerly loads full TE to CUDA and uses `qtype_te`
- [x] Run a full training smoke test on a low-VRAM GPU and confirm model loads and begins training


inflamously:
Omg yes please, anything that fixes these annoying VRAM spikes between sampling and at random steps. I have a 5090 RTX that can run this flux 2 klein 9b without any problems but the longer the run the more it spikes.

jaretburkett:
Thank you @mhofer1976

PR #727, opened 2026-02-27
feat: enable immediate saving on UI stop and enhance optimizer state backup

avan06:
This PR improves the training interruption workflow by ensuring that training progress is captured immediately when a user requests a stop via the UI. It also introduces a backup mechanism for the optimizer state to provide users with more flexibility when resuming or rolling back training.

### Key Changes

1. Immediate Save on UI Stop (DiffusionTrainer)
Modified maybe_stop to trigger self.save() as soon as a stop signal is detected.
Added a _is_saving flag to manage the saving state, preventing infinite recursion or redundant calls during emergency saves.
Benefit: Users can now resume training from the precise step of interruption rather than being forced to roll back to the last scheduled checkpoint. This ensures LoRA weights, Metadata, and optimizer states are perfectly synchronized at the exit point.

2. Optimizer State Rotation (BaseSDTrainProcess)
Implemented a backup system for the optimizer state. When saving, the existing optimizer.pt is moved to optimizer_prev.pt instead of being directly overwritten by the new state.

### Example Scenario (How to roll back)

If a user decides they prefer the results from a previous scheduled save over the final interrupted save:

1. Current State: You have a scheduled save at Step 500 (train_000000500.safetensors and optimizer_prev.pt) and an interruption save at Step 666 (train_000000666.safetensors and optimizer.pt).
2. Rollback Process:
 - Delete the interruption files: train_000000666.safetensors and optimizer.pt.
 - Rename optimizer_prev.pt to optimizer.pt.
3. Result: The trainer will successfully resume training from Step 500 using the correct historical optimizer state.

___

These changes have been verified in a local environment and are confirmed to be working as intended.

### DIFF
--- a/extensions_built_in/sd_trainer/DiffusionTrainer.py
+++ b/extensions_built_in/sd_trainer/DiffusionTrainer.py
@@ -30,6 +30,7 @@ class DiffusionTrainer(SDTrainer):
         
         if self.is_ui_trainer:
             self.is_stopping = False
+            self._is_saving = False
             # Create a thread pool for database operations
             self.thread_pool = concurrent.futures.ThreadPoolExecutor(max_workers=1)
             # Track all async tasks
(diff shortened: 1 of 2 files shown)


hexvalid:
great feature

mcDandy:
Would be usefull when training large models (when it gets stuck at step 480 becouse you did not offload enough) and you do not know if waiting will be better then losing nearly hour of training.

PR #738, opened 2026-03-13
Fix random_noise_multiplier never being applied to the noise.

lRemixl:
The random noise multiplier calculated when self.train_config.random_noise_multiplier > 0.0 is never applied to the noise. This fixes that.


jaretburkett:
Thank you

PR #749, opened 2026-03-23
Add 1328 native resolution for Qwen Image training

Rasaboun:
## Summary

Qwen-Image and Qwen-Image-2512 have a native 1:1 resolution of **1328x1328** as documented in the [official model card](https://huggingface.co/Qwen/Qwen-Image-2512).

- Added 1328 as a resolution option in the training UI


jaretburkett:
Thank you

Issue #752, opened 2026-03-25
I encountered the following error when training a LoRa dataset using "flux2 klein 9b base" on "runpod"


iamddtla0620:
I encountered the following error when training a LoRa dataset using "flux2 klein 9b base" on "runpod":

Error running job: 'AutoEncoder' object has no attribute 'config'

if hasattr(self.sd.vae.config, 'scale_factor_temporal'):

nobandegani:
same here

nobandegani:
this commit is the issue
https://github.com/ostris/ai-toolkit/commit/7f3309b29114a2f47c65d25cfeeb697a0617199f

quick fix for it, change line 511 in ai-toolkit\app\toolkit\data_loader.py
if self.sd is not None:
    if hasattr(self.sd.vae, 'config') and hasattr(self.sd.vae.config, 'scale_factor_temporal'):
        temporal_compression = self.sd.vae.config.scale_factor_temporal
    if hasattr(self.sd.unet, 'config') and hasattr(self.sd.unet.config, 'temporal_compression_ratio'):
        temporal_compression = self.sd.unet.config.temporal_compression_ratio

or if u using docker image, u can use older versions instead of latest tag (ostris/aitoolkit:latest)
https://hub.docker.com/r/ostris/aitoolkit/tags

rfreis:
@nobandegani I pushed a PR with your suggestion as fix

jaretburkett:
Merged @rfreis PR. Sorry about that. I forgot the VAE for flux 2 models doesn't have a config.  https://github.com/ostris/ai-toolkit/pull/754

PR #754, opened 2026-03-25
Fix the bug in temporal_compression data loader

rfreis:
This attempts to fix the issue described on https://github.com/ostris/ai-toolkit/issues/752


rfreis:
@jaretburkett hi man, thanks for your work with AI Toolkit!
I'm not sure if this is what you aim as fix, but seems to be a valid attempt at least.

Can you review and give a feedback?

jaretburkett:
Thank you. I forgot flux2 vae doesn't have config.

PR #766, opened 2026-03-29
Push intermediate checkpoints to HuggingFace Hub

Rasaboun:
## Summary

Adds a `push_to_hub_every_save` option so intermediate checkpoints get pushed to HuggingFace Hub as they are saved, not just at the end of training.

Right now if training crashes before completion you lose everything because Hub push only happens after the loop finishes. With this option enabled each checkpoint saved during training also gets uploaded.

Also wraps the end-of-training Hub push in a try/except so a failed upload no longer crashes the process — the model is already saved locally at that point.

## Scope

Applies to all training processes that extend `BaseSDTrainProcess` (LoRA, Slider, SD Rescale, etc.). VAE and ESRGAN training extend `BaseTrainProcess` directly and do not have Hub push support — that is pre-existing and unchanged by this PR.

## Usage

```yaml
save:
  push_to_hub: true
  push_to_hub_every_save: true
  hf_repo_id: your-username/your-model
```

Defaults to `false` so existing behavior is unchanged. Requires `push_to_hub: true` and `HF_TOKEN` in the environment (no interactive login prompt during training).

## Details

- Uses the existing `post_save_hook` extension point in `BaseSDTrainProcess` which runs after every checkpoint save
- Mid-training and end-of-training Hub pushes are wrapped in try/except to avoid crashing on upload failures
- Adds the field to the UI TypeScript types and default job config
- Documents the option in all example YAML configs

### DIFF
--- a/config/examples/train_lora_chroma_24gb.yaml
+++ b/config/examples/train_lora_chroma_24gb.yaml
@@ -22,7 +22,8 @@ config:
         save_every: 250 # save every this many steps
         max_step_saves_to_keep: 4 # how many intermittent saves to keep
         push_to_hub: false #change this to True to push your trained model to Hugging Face.
-        # You can either set up a HF_TOKEN env variable or you'll be prompted to log-in         
+        push_to_hub_every_save: false # push to HuggingFace Hub every time a checkpoint is saved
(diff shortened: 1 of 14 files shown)


Rasaboun:
@jaretburkett What do you think about this feature ?

PR #805, opened 2026-04-26
Add Rose optimizer

MatthewK78:
Rose optimizer: https://github.com/MatthewK78/Rose

- Stateless: uses less memory than even 8-bit versions of other optimizers. If not for temporary working memory, its memory usage would be as low as plain vanilla SGD (**without** momentum).
- Matches or exceeds AdamW in observed accuracy and convergence.
- Training loss runs higher while validation loss runs lower, consistent with implicit regularization rather than memorization.

Note: Requires a higher than usual learning rate. For example, 1e-3 instead of AdamW's 1e-4.

### DIFF
--- a/requirements_base.txt
+++ b/requirements_base.txt
@@ -41,3 +41,4 @@ av==16.0.1
 torchcodec==0.9.1
 librosa==0.11.0
 mutagen==1.47.0
+rose-opt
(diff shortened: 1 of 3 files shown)


funwithforks:
This optimizer is quite good.

PR #808, opened 2026-04-28
RLHF Flow-GRPO implementation POC

ifilipis:
Added reinforcement learning (Flow-GRPO) that seems to work quite universally across models.

It implements [Flow-GRPO](https://github.com/yifan123/flow_grpo) and lets you vote live and thus create a LoRA

Default parameters seem quite well tuned for quick results within a few iterations. 
For extra fast results, set LR to 0.005 and behold the speed
The only difference in this implementation vs original: rewards are binary instead of relying on a ranking model

There's a new job type dropdown for creating Flow-GRPO tasks, and GRPO job has a voting interface that lets you generate samples and vote on them

Stuff yet to do: 

- [x] Manual checkpoints
- [x] Reduce memory usage (Z-Image now takes 6.5GB in uint4) and improve speed
- [x] UI polishing
- [x]    - Prevent prompt settings from resetting on leaving the tab
- [x]    - Fix skip button color when selected
- [x]    - Add fixed/random seed
- [x]    - Improve progress indication when generating candidates
~~- Maybe fix concurrency of training/generation~~
- [x]    - Improve candidate images layout
- [x]    - Add input images for editing models
- [ ] Keep testing the algorithm on all models


xd2333:
Can it support the training of the editing model? (Add control image and caption)

ifilipis:
> Can it support the training of the editing model? (Add control image and caption)

Reference latents are implemented, but not added to the UI yet

georgeYanch:
Hi, thank you for such a cool feature!

I tried to finetune my Flux Klein Base 9B lora for image editing task, but i see that applying vote takes ~20min on RTX4090 48GB, is this an expected behaviour?

ifilipis:
Yeah, this is expected. With 4 candidates and 30 steps, the trainer actually makes 120 timesteps. What you can do for Flux is take this https://huggingface.co/bghira/flux2-klein-9b-distillation-adapter and add it as both assistant lora and inference lora in config. Then you can reduce generation step count to around 10 or sometimes less. The model will slightly drift after RLHF (it won't be visible in comfyui though), so the best I found is to do another 50-100 regular lora steps after this manipulation

ifilipis:
Added Diffusion-KTO https://arxiv.org/pdf/2404.04465 with both online and offline modes. It is much faster than GRPO, can work with single-image rollouts, but requires more votes and steps overall

In dataset mode, the expected format is two subfolders /pos and /neg - upvotes go into /pos, downvotes go into /neg.
It can also read prompts from ComfyUI metadata. The idea is to generate batch images using ComfyUI, then sort them into folders - the trainer will handle the rest.

References are not implemented yet. It will be merged here as soon as those are done

https://github.com/ifilipis/ai-toolkit/tree/rlhf-kto

Issue #758, opened 2026-03-26
LTX‑2 LoRA training is ~5× slower in AI‑Toolkit compared to LTX Trainer with compiled Triton 3.4.0 (Windows)


zappazack:
Summary  
When training an LTX‑2 LoRA on Windows with a 60‑image high‑quality dataset, AI‑Toolkit runs approximately 5× slower than the official LTX community trainer (1:49h vs. 4:24h (canceled)).
The community trainer uses a self‑compiled Triton 3.4.0 build that works correctly under Windows and achieves full GPU utilization.
AI‑Toolkit, however, shows very low GPU load and significantly longer step times.

Environment
Windows 11

NVIDIA RTX 5090 (32 GB VRAM)

CUDA 13.0

Python 3.13.6

Self‑compiled Triton 3.4.0 (fully functional with LTX trainer)

AI‑Toolkit (latest version)

LTX‑2.3‑22B‑dev model

Dataset: 60 high‑quality images

Observed Behavior
AI‑Toolkit training speed: ~5.4 seconds per step

LTX community trainer speed: ~1.1 seconds per step

GPU utilization during AI‑Toolkit training: ~20–25%

GPU utilization during LTX trainer: ~90–100%

Power draw in AI‑Toolkit: ~160 W

Power draw in LTX trainer: ~350–450 W

Both tests used the same dataset and similar LoRA settings.

Expected Behavior
Significantly higher GPU utilization and training speed closer to the LTX community trainer, or at least performance scaling that reflects the hardware capabilities of the RTX 5090.

jaretburkett:
It could be a number of issues, please post your config file.

jono0301:
I was able to speed up training significantly by adding these DataLoader optimizations under the train section:

train:
  dataloader_num_workers: 8
  prefetch_factor: 8
  pin_memory: true

This resolved a CPU/IO bottleneck where the GPU was waiting for data between batches. After adding these, my iterations per second increased from ~0.005 to ~0.17 (34x speedup), and ETA dropped significantly.

Note: This was with a pure video dataset (121 frames per sample). Not sure if that context matters, but wanted to mention it in case it affects how these parameters behave.

zappazack:
> I was able to speed up training significantly by adding these DataLoader optimizations under the train section:
> 
> train: dataloader_num_workers: 8 prefetch_factor: 8 pin_memory: true
> 
[...]

PR #813, opened 2026-05-01
perf: enable pin_memory on CUDA DataLoaders for faster CPU->GPU transfer

Booyaka101:
## Summary

Adds an opt-in `pin_memory` field on `DatasetConfig` (default `False`). When `True` and CUDA is available, sets `pin_memory=True` on the DataLoader for faster CPU→GPU transfer.

## Empirical validation (RTX 5090 / Ryzen 9 9900X / 22B LTX-2.3 LoRA)

@wobba's measurements [in #758](https://github.com/ostris/ai-toolkit/issues/758#issuecomment-4373355355):

| Config                                                  | Step time | VRAM            |
|---------------------------------------------------------|-----------|-----------------|
| baseline (`low_vram`, `layer_offloading_transformer_percent: 1.0`) | ~20 s     | ceiling, thrashing |
| **+ this PR's `pin_memory: true`**                      | **10 s**  | ~18 GB          |
| + `layer_offloading_transformer_percent: 0.5`           | 6.5 s     | 26.7 GB         |
| + `layer_offloading_transformer_percent: 0.25`          | 4.6 s     | 31 GB           |

The first row→second row delta is a clean isolate of `pin_memory: true` (no other variables changed): **2× speedup**. The remaining rows are independent (the `layer_offloading_transformer_percent` knob), but `pin_memory: true` is what makes those further reductions safe by cleaning up the CPU→GPU transfer path so the user can shift more blocks resident.

## History

The original PR shipped this unconditionally on CUDA, framed as a strict-improvement. @zappazack tested empirically on RTX 5090 + Qwen LoRA at 1024×1024 and found a real regression: when VRAM saturated (~31.3 / 32 GB), step times went from 1.5 s to 20+ s as NVIDIA's Windows driver tried to fall back to shared-memory VRAM overflow but couldn't relocate page-locked pages. PCIe thrashing.

So `pin_memory` is **not** a strict improvement on Windows — it has a real downside for users at the VRAM ceiling. This update converts it to opt-in. @wobba's data above confirms default-false is the right choice (their own recipe explicitly says "back off if step time suddenly spikes").

## Mechanism (for posterity)

`pin_memory=True` allocates batch tensors in page-locked system RAM so PyTorch can issue async host→device DMA copies. Win is real on stable-VRAM systems (typical CPU-bound DataLoader pipelines).

But on Windows with NVIDIA driver shared-memory VRAM overflow active (default in modern drivers), the driver tries to use system RAM as a VRAM-overflow buffer when GPU memory saturates. Page-locked pages cannot be relocated. The marginal pinned RAM from this PR (~100 MB of DataLoader batch tensors) is small in absolute terms but enough to push thrashing onset earlier when the user is right at the VRAM ceiling.

Note: ai-toolkit's existing `layer_offloading` system already pins multi-GB of weight tensors via `toolkit/memory_management/manager_modules.py:111` (`t.pin_memory()`). The same shared-memory-fallback conflict applies there too — this PR just exposes the knob to users for the DataLoader-level pin specifically.

## Diff

```diff
# toolkit/config_modules.py — DatasetConfig
+ # Pin DataLoader output tensors in page-locked RAM for faster CPU->GPU
+ # transfer. Off by default because page-locked RAM cannot be relocated
+ # by NVIDIA's Windows driver shared-memory VRAM-overflow fallback,
+ # which can cause severe PCIe thrashing for users at the VRAM ceiling.
+ # Opt in if you have stable VRAM headroom and want the transfer speedup.
+ self.pin_memory: bool = kwargs.get('pin_memory', False)

# toolkit/data_loader.py — get_dataloader_from_datasets
+ if torch.cuda.is_available() and dataset_config_list[0].pin_memory:
+     dataloader_kwargs['pin_memory'] = True
```

14 lines total across two files. No API change beyond the new optional config field.

## Usage

In dataset config:

```yaml
datasets:
  - folder_path: "..."
    pin_memory: true   # opt-in if you have stable VRAM headroom
```

## Refs

Closes the regression @zappazack reported in #758. Empirical numbers from @wobba on the same thread. The bigger Windows perf gap (`num_workers=0` constraint due to unpickleable `self.sd` on the dataset) is a separate concern surfaced in the same comment thread — orthogonal to this PR.


Booyaka101:
Updated. `pin_memory` is now opt-in via dataset config (default `False`), gated on `torch.cuda.is_available() AND dataset_config_list[0].pin_memory`. Same 7-line shape, just one extra config field on `DatasetConfig` and one extra condition.

@zappazack — your Qwen 5090 setup will be unaffected by this PR landing, since the default is off. Users who explicitly opt in (with `pin_memory: true` in dataset config) accept the tradeoff. Updated commit: 0c8793f. Thanks again for the empirical catch.

crashless:
I just wanted to chime in here and support the findings. rtx5090 training LTX2.3, 96gb system ram on windows. 

I implemented #813 and training went from ~20s per step to ~10.5s per step. I also see a dramatic increase in effective power usage from ~150w up to 250-300w. This is without optimizing the offload %. 

Clearly this pin_memory flag makes a big difference. I didn't do a comprehensive VRAM comparison as both with and without pin_memory are using considerable "shared gpu memory" (5-12gb+ overflow). However whatever magic this is doing clearly is improving performance dramatically. 

Nice work - #813 should be implemented asap.

wobba:
@jaretburkett do you see a reason to re-open this one? Or addressed in other ways. Maybe we need to auto-post comment to avoid auto-close ;)

jaretburkett:
Thank you

PR #829, opened 2026-05-11
Add dual-GPU model-parallel paths for 8 LoRA-training architectures (FLUX.2 / Wan 2.2 / LTX-2 / Qwen-Image / Chroma / FLUX.1-Kontext / HiDream-I1 / Z-Image)

### LINKED ISSUE #531: Flux.2-dev training does not work on my big computer
## This is for bugs only

Did you already ask [in the discord](https://discord.gg/VXmU2f5WEU)?

Yes/No

You verified that this is a bug and not a feature request or question by asking [in the discord](https://discord.gg/VXmU2f5WEU)?

Yes/No

## Describe the bug

For some reasons
- The training eats up 140 GB/192 GB of my ram.
- It fills my whole 5090's VRAM (I have two 5090s but the script won't use the second one. This is not a problem but it's just to say that I have more than enough compute power for the training to work.)
- The VRAM seems to not be enough given how it eats up the shared memory too.
- My computer lags like crazy even to type this text.

It has been one hour that my GPU has been generating the first 1024x1024 image pre-training, and it's still not done. I don't know if it'll be done somedays.

I checked
- low vram
- float8 quantization (I let the quantization options by default)
- cache latents to disk
- I allowed bucket size of 1280 and 1536

My dataset contains 141 labeled images.

image

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 581.80                 Driver Version: 581.80         CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                  Driver-Model | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 5090      WDDM  |   00000000:01:00.0  On |                  N/A |
|  0%   29C    P1            111W /  575W |   31916MiB /  32607MiB |    100%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA GeForce RTX 5090      WDDM  |   00000000:03:00.0 Off |                  N/A |
|  0%   25C    P8             12W /  575W |     837MiB /  32607MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
```
Given how the power stays low even with 100% usage, I suspect that most of the GPU's usage are PCIE communications (the VRAM is overflowing)

Here is the job config to reproduce.
```
{
  "job": "extension",
  "config": {
    "name": "test",
    "process": [
      {
        "type": "diffusion_trainer",
        "training_folder": "C:\\Users\\asuka\\sc\\ia\\AI-Toolkit-Easy-Install\\AI-Toolkit\\output",
        "sqlite_db_path": "C:\\Users\\asuka\\sc\\ia\\AI-Toolkit-Easy-Install\\AI-Toolkit\\aitk_db.db",
        "device": "cuda",
        "trigger_word": null,
        "performance_log_every": 10,
        "network": {
          "type": "lora",
          "linear": 32,
          "linear_alpha": 32,
          "conv": 16,
          "conv_alpha": 16,
          "lokr_full_rank": true,
          "lokr_factor": -1,
          "network_kwargs": {
            "ignore_if_contains": []
          }
        },
        "save": {
          "dtype": "bf16",
          "save_every": 250,
          "max_step_saves_to_keep": 4,
          "save_format": "diffusers",
          "push_to_hub": false
        },
        "datasets": [
          {
            "folder_path": "C:\\Users\\asuka\\sc\\ia\\AI-Toolkit-Easy-Install\\AI-Toolkit\\datasets/test",
            "mask_path": null,
            "mask_min_value": 0.1,
            "default_caption": "",
            "caption_ext": "txt",
            "caption_dropout_rate": 0.05,
            "cache_latents_to_disk": true,
            "is_reg": false,
            "network_weight": 1,
            "resolution": [
              768,
              1024,
              1280,
              1536,
              512
            ],
            "controls": [],
            "shrink_video_to_frames": true,
            "num_frames": 1,
            "do_i2v": true,
            "flip_x": false,
            "flip_y": false,
            "control_path_1": null,
            "control_path_2": null,
            "control_path_3": null
          }
        ],
        "train": {
          "batch_size": 1,
          "bypass_guidance_embedding": false,
          "steps": 3000,
          "gradient_accumulation": 1,
          "train_unet": true,
          "train_text_encoder": false,
          "gradient_checkpointing": true,
          "noise_scheduler": "flowmatch",
          "optimizer": "adamw8bit",
          "timestep_type": "weighted",
          "content_or_style": "balanced",
          "optimizer_params": {
            "weight_decay": 0.0001
          },
          "unload_text_encoder": false,
          "cache_text_embeddings": false,
          "lr": 0.0001,
          "ema_config": {
            "use_ema": false,
            "ema_decay": 0.99
          },
          "skip_first_sample": false,
          "force_first_sample": false,
          "disable_sampling": false,
          "dtype": "bf16",
          "diff_output_preservation": false,
          "diff_output_preservation_multiplier": 1,
          "diff_output_preservation_class": "person",
          "switch_boundary_every": 1,
          "loss_type": "mse"
        },
        "model": {
          "name_or_path": "black-forest-labs/FLUX.2-dev",
          "quantize": true,
          "qtype": "qfloat8",
          "quantize_te": true,
          "qtype_te": "qfloat8",
          "arch": "flux2",
          "low_vram": true,
          "model_kwargs": {
            "match_target_res": false
          },
          "layer_offloading": false,
          "layer_offloading_text_encoder_percent": 1,
          "layer_offloading_transformer_percent": 1
        },
        "sample": {
          "sampler": "flowmatch",
          "sample_every": 250,
          "width": 1024,
          "height": 1024,
          "samples": [
            {
              "prompt": "A cat."
            }
          ],
          "neg": "",
          "seed": 42,
          "walk_seed": true,
          "guidance_scale": 4,
          "sample_steps": 25,
          "num_frames": 1,
          "fps": 1
        }
      }
    ]
  },
  "meta": {
    "name": "[name]",
    "version": "1.0"
  }
}
```

genno-whittlery:
Dual RTX 5090 is the best consumer GPU setup one can get. This PR
proposes a way past the per-card 32 GB VRAM limit for **FLUX.2**,
**Wan 2.2**, **LTX-2**, **Qwen-Image**, **Chroma1-HD**, **FLUX.1-Kontext**,
**HiDream-I1**, and **Z-Image** LoRA training.

Closes #531.

> **Scope note:** originally filed FLUX.2-only; the seven others are
> folded in because they share the same toolkit-wide infrastructure (the
> LoRA per-layer device routing patches, the CPU-quantize-before-move
> load path, the gradient-checkpoint dict/list aliasing handling). The
> dual-GPU machinery is genuinely shared; one PR per trainer is the
> honest unit of work. Happy to split if you'd prefer to review
> separately.

## Problem

On a single 32 GB consumer GPU, FLUX.2-dev at qfloat8 (~30 GB) plus
bf16 activations overshoots VRAM by ~1 GB per forward. Windows WDDM
pages the transient through unified memory every step. Throughput on
a single 5090 collapses from 14.4 s/it to 277 s/it within a few
hundred steps. On Linux the same configuration OOMs outright.

`torch.cuda.memory_summary()` confirms: at step 1 the peak allocation
already exceeds the card by ~1 GB (`cur=30.63 GB / peak=32.94 GB /
capacity=31.84 GB`). It's a per-step over-capacity transient, not
a leak.

The Wan 2.2 14B A14B variants have the same shape of problem — each of
the two expert transformers is ~14 GB at fp8, and the activation
footprint at video resolutions pushes a single 32 GB card over.
LTX-2 / Qwen-Image / Chroma1-HD / FLUX.1-Kontext / HiDream-I1 each hit
the same wall at fp8 + bf16 activations on a single 32 GB card.
Z-Image is the smallest of the eight (~30-block transformer) and would
fit single-card, but the same mixin runs it for symmetry + faster
training at higher batch sizes.

## Fix

Opt-in per arch via `FLUX2_DUAL_GPU=true` / `WAN_DUAL_GPU=true` /
`LTX2_DUAL_GPU=true` / `QWEN_IMAGE_DUAL_GPU=true` / `CHROMA_DUAL_GPU=true`
/ `FLUX_KONTEXT_DUAL_GPU=true` / `HIDREAM_DUAL_GPU=true` /
`Z_IMAGE_DUAL_GPU=true`.

- Transformer split across two CUDA devices at the block-list
  midpoint. One PCIe boundary per forward.
- Text encoder pinned to CPU (`<ARCH>_TE_DEVICE=cpu`). With the existing
  `cache_text_embeddings` + `unload_text_encoder` flags it runs once at
  training start, caches embeddings, then unloads. The inner loop never
  touches it.
- LoRA per-layer device routing. `ToolkitNetworkMixin.force_to` and
  `LoRANetwork.apply_to` patched to follow each LoRA's wrapped layer;
  lazy forward-pin as a safety net for LoRAs added post-install. These
  patches target ai-toolkit's own classes (not arch-specific), so every
  port reuses them via per-class idempotency flags.
- Fails fast on misconfiguration — fewer than 2 CUDA devices, a
  non-CUDA device, or an out-of-range split index each raise a clear
  error at load time rather than crashing mid-forward.

### FLUX.2

Packaged as `Flux2DualGPUMixin` in
`extensions_built_in/diffusion_models/flux2/flux2_dual_gpu.py`;
`Flux2Model` inherits it. Split at the `single_blocks` midpoint —
DoubleStreamBlocks + first half of SingleStreamBlocks on the primary
device, the remainder + `final_layer` on the next ordinal.

### Wan 2.2

Packaged as `Wan22DualGPUMixin` in
`extensions_built_in/diffusion_models/wan22/wan22_dual_gpu.py`, mirroring
the FLUX.2 mixin shape. `Wan225bModel` and `Wan2214bModel` inherit it
(the I2V variant transitively). Split places `rope` / `patch_embedding`
/ `condition_embedder` / first half of `blocks` on cuda:0, the remainder
+ `norm_out` / `proj_out` / `scale_shift_table` on cuda:1.
`Wan21.load_wan_transformer`'s existing
`raise ValueError("Splitting model over gpus is not supported")` is
softened to dispatch to the mixin when `WAN_DUAL_GPU=true`. For the 14B
A14B dual-DiT variants, each of the two expert transformers
(`transformer_1` high-noise + `transformer_2` low-noise) is distributed
individually after quantize; the `DualWanTransformer3DModel` wrapper's
timestep-routed forward is unchanged — whichever expert it picks now
runs through the split forward. Wan-specific handling: `rotary_emb` is a
`(cos, sin)` tuple, `scale_shift_table` is an `nn.Parameter`, and
Wan 2.2 ti2v's 2D-timestep path is preserved in the split forward.

### LTX-2

Packaged as `Ltx2DualGPUMixin` in
`extensions_built_in/diffusion_models/ltx2/ltx2_dual_gpu.py`. 48-block
transformer, split midpoint. LTX-2-specific handling: dual-modality
(video + audio) carries four RoPE tuples through the boundary; the
post-split forward bridges all four. The audio TE / connectors / vocoder
are kept on `te_device_torch` alongside the main TE.

### Qwen-Image

Packaged as `QwenImageDualGPUMixin` in
`extensions_built_in/diffusion_models/qwen_image/qwen_image_dual_gpu.py`.
60-block transformer, split 30/30. The split forward bridges
`hidden_states`, `encoder_hidden_states`, `temb`, the `(img_freqs,
txt_freqs)` RoPE tuple, the joint attention mask, and `modulate_index`
at the boundary. Two notable bugs surfaced + fixed:

  - The recreated split forward was originally written against the
    mac-side diffusers `QwenImageTransformer2DModel.forward`, which
    lacks the deprecated `txt_seq_lens` arg present on the box.
    Recreated against the box's copy so the signature + body match.
  - Gradient-checkpoint dict aliasing — `block_attention_kwargs` is a
    dict captured by reference per block; mutating its
    `"attention_mask"` entry at the boundary corrupted the cuda:0
    blocks' backward recompute. Fix: build a *separate* cuda:1 dict at
    the boundary; the cuda:0 dict is read-only post-distribute.

Four hardcoded `device_torch` text-encoder sites in `qwen_image.py` are
routed to `te_device_torch` so the Qwen2.5-VL TE lands on CPU under
dual-GPU.

### Chroma1-HD

Packaged as `ChromaDualGPUMixin` in
`extensions_built_in/diffusion_models/chroma/chroma_dual_gpu.py`. Chroma
is FLUX-shaped (19 double + 38 single blocks + final_layer) on the
*vendored* `Chroma` module under `chroma/src/model.py` — NOT a diffusers
class. The trainer imports the in-tree `src/model.py` via
`from .src.model import Chroma`, so that is the authoritative
split-forward source.

Chroma's `distilled_guidance_layer` runs under `no_grad` on cuda:0 once
before the block loops, producing `mod_vectors` distributed into a dict
keyed by block name (`single_blocks.{i}.modulation.lin`,
`double_blocks.{i}.img_mod.lin` / `.txt_mod.lin`,
`final_layer.adaLN_modulation.1`). At the PCIe boundary the forward
builds a *separate* cuda:1-resident dict of `ModulationOut` dataclasses
for the cuda:1 single blocks plus the final-layer modulation tensors —
the Qwen-Image gradient-checkpoint dict-aliasing lesson generalized to
this dict shape. The `mod_index` buffer (registered non-persistent on
the parent module) gets an explicit `.to(d0)` since per-submodule `.to()`
doesn't reach module-level buffers. Chroma's CLIP slot is a `FakeCLIP`
stub; only T5 is real and routes to `te_device_torch`.

### FLUX.1-Kontext

Packaged as `FluxKontextDualGPUMixin` in
`extensions_built_in/diffusion_models/flux_kontext/flux_kontext_dual_gpu.py`.
Targets the diffusers `FluxTransformer2DModel` (distinct from
ai-toolkit's vendored Flux2 transformer — different module attr names:
`transformer_blocks` / `single_transformer_blocks` / `proj_out` instead
of `double_blocks` / `single_blocks` / `final_layer`). Split places
`x_embedder` / `context_embedder` / `time_text_embed` / `pos_embed` +
all 19 `transformer_blocks` + `single_transformer_blocks[:19]` on
cuda:0; `single_transformer_blocks[19:]` + `norm_out` + `proj_out` on
cuda:1. Boundary bridges `hidden_states`, `encoder_hidden_states` (both
block loops return BOTH after the recent diffusers refactor), `temb`,
the `image_rotary_emb` (cos, sin) tuple, and any tensor entries in
`joint_attention_kwargs` (built as a NEW dict so the cuda:0 view stays
untouched). Controlnet residual branches preserved verbatim from the
upstream forward for parity.

Bug surfaced + fixed: `get_prompt_embeds` force-moved CLIP-L to
`self.device_torch` (cuda:0) while T5 stayed on cpu under
`FLUX_KONTEXT_TE_DEVICE=cpu`. The shared
`train_tools.encode_prompts_flux` helper then reads
`device = text_encoder[0].device` (CLIP-L → cuda:0) and piped cuda:0
`input_ids` to the cpu-resident T5 — token_embedding crashed
"index on cuda:0, weight on cpu". Fix: route both TEs through
`te_device_torch` in `get_prompt_embeds`, then move the resulting
`prompt_embeds + pooled_prompt_embeds` back to `device_torch` for
downstream pipeline consumption.

### HiDream-I1

Packaged as `HiDreamDualGPUMixin` in
`extensions_built_in/diffusion_models/hidream/hidream_dual_gpu.py`.
Targets the *vendored* `HiDreamImageTransformer2DModel` under
`hidream/src/models/transformers/transformer_hidream_image.py` — NOT
the diffusers-package class, which has a substantially different
forward (~1024 lines of divergence). The hidream extension imports from
the vendored copy via `from .src.models.transformers.transformer_hidream_image
import HiDreamImageTransformer2DModel`, so that is the authoritative
split-forward source.

Split places `x_embedder` / `t_embedder` / `p_embedder` / `pe_embedder` /
`caption_projection` + ALL `double_stream_blocks` (16) +
`single_stream_blocks[:16]` on cuda:0; `single_stream_blocks[16:]` +
`final_layer` on cuda:1. HiDream's `_no_split_modules` attribute governs
accelerate device-map sharding (it forbids splitting *within* a block),
and our split is between blocks, so it's informational only. MoE
(`MoEGate` / `MOEFeedForwardSwiGLU`) lives inside the blocks and
distributes with them transparently.

Bug surfaced + fixed: `get_prompt_embeds` passed `device=self.device_torch`
to the pipeline's `_encode_prompt`, which routed `input_ids` to cuda:0
while the four TEs lived on cpu under `HIDREAM_TE_DEVICE=cpu`. CLIP's
token_embedding crashed "index on cuda:0, weight on cpu". Fix: route
the encode through `te_device_torch`, then recursively move the
returned shapes (HiDream's `_encode_prompt` returns `prompt_embeds` as a
*list* of per-encoder tensors plus a single `pooled_prompt_embeds`
tensor) back to `device_torch` via a `_to_device` helper that handles
both list and scalar-tensor outputs. Te_device_torch is load-bearing
for the ~30 GB Llama-3.1-8B encoder — pinning it to CPU keeps cuda:0
free for the distributed transformer + activations.

### Z-Image

Packaged as `ZImageDualGPUMixin` in
`extensions_built_in/diffusion_models/z_image/z_image_dual_gpu.py`.
Targets the diffusers `ZImageTransformer2DModel`, a novel architecture
distinct from the FLUX-family shape every prior port used:

  - `all_x_embedder` / `all_final_layer`: `nn.ModuleDict` keyed by
    `"{patch_size}-{f_patch_size}"`
  - `noise_refiner` / `context_refiner` / optional `siglip_refiner`:
    `nn.ModuleList` (2 blocks each)
  - `x_pad_token` / `cap_pad_token` / optional `siglip_pad_token`:
    *module-level `nn.Parameter` attributes* — don't follow per-
    submodule `.to()` calls; require explicit `.data = .data.to(d0,
    dtype=dtype)` moves
  - `layers`: `nn.ModuleList` of 30 main blocks (the split target)
  - `rope_embedder`: a plain Python class (NOT `nn.Module` — auto-
    follows input device on each call)

Split places every pre-main-loop module (the four embedders, both
refiners, the pad-token Parameters) plus `layers[:15]` on cuda:0;
`layers[15:]` + `all_final_layer` on cuda:1.

Z-Image's forward takes list-style inputs (`x: list[Tensor]`,
`cap_feats: list[Tensor]`) rather than packed tensors. The split
forward runs through `patchify_and_embed` → x embed/refine → cap
embed/refine → optional siglip embed/refine → `_build_unified_sequence`
on cuda:0 before the main `self.layers` loop. At the PCIe boundary
inside the main loop it bridges `unified`, `unified_mask`,
`unified_freqs`, `adaln_input` (or `t_noisy` / `t_clean` in Omni mode),
and `unified_noise_tensor` to cuda:1. After `all_final_layer` +
`unpatchify` the `list[Tensor]` output is moved back to cuda:0
element-wise. `get_prompt_embeds` routes through `te_device_torch` so
the Qwen3 encoder stays on CPU through caching.

## Validation

All validated end-to-end on 2× RTX 5090.

**FLUX.2** — character LoRA, FLUX.2-dev + Turbo, rank 16, 512², batch 1,
gradient_checkpointing, adamw8bit, 1500 steps:

| Setup | Step rate | Wall clock (400 steps) |
|---|---|---|
| Single RTX 5090 (current default) | 14.4 → 277 s/it (WDDM-thrash) | 90 min → 30+ hours |
| Dual RTX 5090 (this PR) | 2.85 s/it sustained | ~19 min |

Per-GPU footprint: GPU 0 ~21 GB, GPU 1 ~12 GB. Both cards alternate at
99% SM, single-microbatch pipeline-parallel. Gradient-checkpointing
recompute crosses the device boundary cleanly (non-reentrant), and
`clip_grad_norm_` is compatible with the split.

**FLUX.2-klein** — the klein variants share the architecture and inherit
`Flux2DualGPUMixin` through `Flux2KleinModel`. Both validated, 100-step
smoke, sumi dataset, rank 16, fp8, 512²:

| Variant | Block split | Step rate |
|---|---|---|
| FLUX.2-klein-4B | 10/10 single-blocks | 2.33 it/s |
| FLUX.2-klein-9B | 12/12 single-blocks | 1.21 it/s |

**Wan 2.2** — sumi dataset, rank 16, fp8 weight-only, 512²:

| Variant | Detail | Step rate |
|---|---|---|
| Wan 2.2 TI2V-5B | 50 steps | 1.86 it/s |
| Wan 2.2 T2V-A14B | dual-DiT, `switch_boundary_every=10` (both experts exercised), both transformers distributed across cuda:0/cuda:1 | 1.09 s/it |

**LTX-2** — sumi dataset, rank 16, `--fp8_base`, 512²:

| Variant | Detail | Step rate |
|---|---|---|
| LTX-2 | 48 blocks split 24/24, Gemma3 12B TE on CPU, 20 steps | 2.78 s/it |

**Qwen-Image** — sumi dataset, rank 16, `--fp8_base`, 512²:

| Variant | Detail | Step rate |
|---|---|---|
| Qwen-Image | 60 blocks split 30/30, Qwen2.5-VL TE on CPU, 20 steps | 1.56 s/it |

**Chroma1-HD / FLUX.1-Kontext / HiDream-I1 / Z-Image** — sumi-512 dataset,
rank 16, fp8, 512², 20-step smoke, dual-GPU + TE on CPU:

| Variant | Detail | Step rate | Wall |
|---|---|---|---|
| Chroma1-HD | 19+38 blocks, split 19/19 single, T5 TE on CPU | 1.09 it/s | 4m49s |
| FLUX.1-Kontext | 19+38 blocks (diffusers `FluxTransformer2DModel`), split 19/19 single, CLIP+T5 on CPU | 1.05 s/it | 3m10s |
| HiDream-I1 | 16+32 blocks (vendored transformer), split 16/16 single, 4 TEs on CPU (incl. ~30 GB Llama-3.1-8B) | 2.1 s/it | 5m49s |
| Z-Image | 30 main `layers`, split 15/15, Qwen3 TE on CPU | (fast) | 0m59s |

## New environment variables

- `<ARCH>_DUAL_GPU=true` — enable the dual-GPU path (one per arch:
  FLUX2, WAN, LTX2, QWEN_IMAGE, CHROMA, FLUX_KONTEXT, HIDREAM, Z_IMAGE)
- `<ARCH>_TE_DEVICE=cpu` — pin text encoder to a device (independently
  useful)
- `<ARCH>_DUAL_GPU_SPLIT_AT` — override the block-list split index
  (default: midpoint of whichever block list is the split target)
- `FLUX2_MISTRAL_PATH=/local/path` — skip HF download for Mistral

## Limitations

fp8 weight-only training; pure-bf16 would need a third 32 GB card or
layer offloading. Single-microbatch pipeline-parallel; no gpipe / 1F1B
yet. EMA (`ema_config`) and `low_vram` transformer offloading do not
apply with the split active. Wan 2.2 A14B was validated with both
experts exercised via `switch_boundary_every`; the explicit
`--dit` + `--dit_high_noise` dual-checkpoint path is exercised through
the same `DualWanTransformer3DModel` wrapper. Z-Image's
`layer_offloading` config option is gated off when
`Z_IMAGE_DUAL_GPU=true` (both are competing memory schemes — picking
dual-GPU disables the alternative paths cleanly).

## Companion repo

Standalone version + writeup + diagnostic patches, plus the same
dual-GPU split ported to musubi-tuner, DiffSynth-Studio, and OneTrainer:
https://github.com/genno-whittlery/flux2-dual-gpu-lora

### DIFF
--- a/extensions_built_in/diffusion_models/chroma/chroma_dual_gpu.py
+++ b/extensions_built_in/diffusion_models/chroma/chroma_dual_gpu.py
@@ -1,0 +1,601 @@
+"""Dual-GPU model-parallel path for Chroma1-HD LoRA training.
+
+Activates when ``CHROMA_DUAL_GPU=true``. Distributes the ``Chroma``
+transformer (ai-toolkit's vendored copy under ``chroma/src/model.py`` —
+Chroma is its own implementation, not a diffusers class) across two CUDA
+devices with a single PCIe boundary mid-``single_blocks``, keeps the T5
+text encoder on a configurable device (``CHROMA_TE_DEVICE``, defaulting
+to ``device_torch``), and reuses the same LoRA / multiplier patches the
+FLUX.2 / Wan 2.2 / LTX-2 / Qwen-Image / HiDream dual-GPU paths install
+(those hook ai-toolkit-wide classes, not model-specific ones, so they
+apply unchanged here — the shared per-target patch flags make calling
+them all safe).
+
+Targets the ai-toolkit Chroma trainer
+(``extensions_built_in/diffusion_models/chroma/chroma_model.py``).
+Integration is a small edit at the call site:
+
+1. Make ``ChromaModel`` inherit from ``ChromaDualGPUMixin`` first
+   (MRO ordering: mixin before ``BaseModel``).
+2. In ``ChromaModel.__init__``, call ``self.init_te_device()`` after the
+   base ``__init__`` to resolve ``te_device_torch``.
+3. In ``ChromaModel.load_model``, after quantize, call
+   ``self.setup_dual_gpu_distribution(transformer, dtype)`` gated by
+   ``is_dual_gpu_enabled()``. Route the T5 ``.to()`` sites to
+   ``self.te_device_torch`` instead of ``self.device_torch`` (the
+   FakeCLIP stub site is a no-op and can stay).
+
+Chroma has the FLUX shape: a sequence of ``double_blocks`` (19), then a
+``cat(txt, img)``, then ``single_blocks`` (38), then a strip + ``final_layer``.
+All loop-invariant tensors (``pe``, ``txt_img_mask``, the modulation-vector
+dict produced once by ``distilled_guidance_layer``) are precomputed before
+the block loops, so a single-boundary forward replacement is sufficient.
+
+The split: all 19 ``double_blocks`` plus ``single_blocks[:split_at]`` live
+on cuda:0; the remaining ``single_blocks`` plus ``final_layer`` live on
+cuda:1, with one PCIe boundary inside the single-block loop. Default
+``split_at = num_single_blocks // 2 = 19``.
+
+Env vars:
+    CHROMA_DUAL_GPU=true              enable the dual-GPU path
+    CHROMA_TE_DEVICE=cpu              pin the T5 text encoder
+    CHROMA_DUAL_GPU_SPLIT_AT=19       override single_blocks split index
+                                      (default: num_single_blocks // 2)
+"""
+from __future__ import annotations
+
+import os
+import types
+from typing import TYPE_CHECKING, Any
+
+import torch
+
+from diffusers.utils import logging
+
+if TYPE_CHECKING:
+    from .chroma_model import ChromaModel  # noqa: F401
+
+logger = logging.get_logger(__name__)  # pylint: disable=invalid-name
+
+
+def is_dual_gpu_enabled() -> bool:
+    return os.getenv("CHROMA_DUAL_GPU", "false").lower() == "true"
+
+
+def get_te_device_override() -> torch.device | None:
+    """Returns the text-encoder device override if CHROMA_TE_DEVICE is set."""
+    val = os.getenv("CHROMA_TE_DEVICE")
+    return torch.device(val) if val else None
+
+
+class ChromaDualGPUMixin:
+    """Mixin for :class:`ChromaModel` that adds the dual-GPU training path.
+
+    The mixin contributes:
+
+    - ``te_device_torch`` attribute (resolved from ``CHROMA_TE_DEVICE``
+      or defaulting to ``device_torch``). Always present — used even when
+      dual-GPU itself is disabled.
+    - ``text_encoder_to`` override that routes the T5 text encoder to
+      ``te_device_torch`` instead of ``device_torch``. The CLIP slot in
+      Chroma is a ``FakeCLIP`` stub with no real parameters; iterating
+      the list is harmless.
+    - ``setup_dual_gpu_distribution(transformer, dtype)`` to be called
+      from ``load_model`` after the transformer has been quantized (if
+      applicable) and before the pipeline composes it. Distributes
+      modules across cuda:0/cuda:1, replaces ``Chroma.forward`` with a
+      split-aware variant, pins ``transformer.to()`` to ignore device
+      arguments (preserving the layout against later ai-toolkit ``.to()``
+      calls), and installs the LoRA / multiplier patches.
+    """
+
+    te_device_torch: torch.device
+
+    # ---------------------------------------------------------------- init
+
+    def init_te_device(self) -> None:
+        """Resolve ``te_device_torch`` from env var or fall back to model device.
+
+        Must be called from :meth:`ChromaModel.__init__` after the base
+        ``__init__`` (which establishes ``self.device_torch``).
+        """
+        override = get_te_device_override()
+        self.te_device_torch = override if override is not None else self.device_torch  # type: ignore[attr-defined]
+        if self.te_device_torch.type == "cpu":
+            self.print_and_status_update(  # type: ignore[attr-defined]
+                "CHROMA_TE_DEVICE=cpu — the T5 text encoder will run on CPU. "
+                "Set `cache_text_embeddings: true` in your config so it runs "
+                "once at startup; otherwise it runs on CPU every step."
+            )
+
+    # ----------------------------------------------------------- TE override
+
+    def text_encoder_to(self, *args: Any, **kwargs: Any) -> None:
+        """Override of BaseModel.text_encoder_to that honors te_device_torch.
+
+        ai-toolkit's :class:`SDTrainer` hook ``hook_before_train_loop``
+        calls ``self.sd.text_encoder_to(self.device_torch)``
+        unconditionally. With ``CHROMA_TE_DEVICE=cpu`` we want the T5
+        encoder to stay on CPU regardless. Chroma stores text encoders as
+        a list (``[FakeCLIP, T5]``); iterate it. FakeCLIP's ``.to()`` is
+        benign — it has no real parameters.
+        """
+        target = self.te_device_torch
+        if isinstance(self.text_encoder, list):  # type: ignore[attr-defined]
+            for encoder in self.text_encoder:  # type: ignore[attr-defined]
+                encoder.to(target)
+        else:
+            self.text_encoder.to(target)  # type: ignore[attr-defined]
+
+    # ------------------------------------------------------------ load hook
+
+    def setup_dual_gpu_distribution(self, transformer: torch.nn.Module, dtype: torch.dtype) -> None:
+        """Distribute the Chroma transformer across cuda:0 and cuda:1.
+
+        Layout:
+            cuda:0  — ``img_in``, ``txt_in``, ``pe_embedder``,
+                      ``distilled_guidance_layer``, ALL ``double_blocks``,
+                      ``single_blocks[:split_at]``
+            cuda:1  — ``single_blocks[split_at:]``, ``final_layer``
+
+        ``Chroma`` has one module-level buffer (``mod_index``, non-persistent)
+        which is moved with the parent ``.to(d0, dtype=dtype)`` of the
+        whole transformer at the bottom of this function via the explicit
+        per-submodule ``.to()`` calls — except ``mod_index`` lives directly
+        on the parent module, so we move it explicitly. RoPE frequencies
+        are computed dynamically inside ``EmbedND.forward`` from
+        ``params.theta`` (no plain-attribute freq tensor to move).
+
+        Also installs the LoRA / multiplier patches that align downstream
+        ai-toolkit machinery with the split layout.
+        """
+        if not torch.cuda.is_available() or torch.cuda.device_count() < 2:
+            raise RuntimeError(
+                "CHROMA_DUAL_GPU=true needs at least 2 CUDA devices (cuda:0 + "
+                f"cuda:1); found {torch.cuda.device_count() if torch.cuda.is_available() else 0}. "
+                "Unset CHROMA_DUAL_GPU for single-GPU training."
+            )
+
+        d0 = torch.device("cuda:0")
+        d1 = torch.device("cuda:1")
+        n_single = len(transformer.single_blocks)
+        override = os.getenv("CHROMA_DUAL_GPU_SPLIT_AT")
+        if override is not None:
+            try:
+                split_at = int(override)
+            except ValueError:
+                raise ValueError(
+                    "CHROMA_DUAL_GPU_SPLIT_AT must be an integer; got "
+                    f"{override!r}."
+                )
+            if not (1 <= split_at < n_single):
+                raise ValueError(
+                    f"CHROMA_DUAL_GPU_SPLIT_AT={split_at} is out of range; "
+                    f"must be in [1, {n_single - 1}] for a model with "
+                    f"{n_single} single_blocks."
+                )
+        else:
+            split_at = n_single // 2
+
+        self.print_and_status_update(  # type: ignore[attr-defined]
+            f"Distributing transformer across {d0} and {d1}"
+        )
+        self.print_and_status_update(  # type: ignore[attr-defined]
+            f"Single-block split: {split_at} on {d0}, "
+            f"{n_single - split_at} on {d1}"
+        )
+
+        # --- cuda:0: embedders, distillation approximator, double-stream
+        #     blocks, first half of single-stream blocks
+        transformer.img_in.to(d0, dtype=dtype)
+        transformer.txt_in.to(d0, dtype=dtype)
+        # pe_embedder (EmbedND) has no parameters — RoPE is computed
+        # dynamically via torch.arange / torch.outer inside its forward —
+        # but call .to so child buffers (if any future addition) follow.
+        transformer.pe_embedder.to(d0, dtype=dtype)
+        transformer.distilled_guidance_layer.to(d0, dtype=dtype)
+
+        for blk in transformer.double_blocks:
+            blk.to(d0, dtype=dtype)
+
+        for i, blk in enumerate(transformer.single_blocks):
+            blk.to(d0 if i < split_at else d1, dtype=dtype)
+
+        # --- cuda:1: final output layer
+        transformer.final_layer.to(d1, dtype=dtype)
+
+        # mod_index buffer (registered non-persistent on the parent) feeds
+        # timestep_embedding under no_grad on cuda:0; keep it on d0.
+        if hasattr(transformer, "mod_index"):
+            transformer.mod_index = transformer.mod_index.to(d0)
+
+        transformer.forward = types.MethodType(
+            _make_split_forward(split_at), transformer
+        )
+        _pin_transformer_to(transformer)
+
+        _install_external_patches()
+
+
+# ---------------------------------------------------------------- helpers
+
+
+def _make_split_forward(split_at: int):
+    """Create a ``Chroma.forward`` variant that inserts a cuda:1 boundary
+    at ``single_blocks[split_at]``.
+
+    Recreates ai-toolkit's vendored ``Chroma.forward``
+    (``extensions_built_in/diffusion_models/chroma/src/model.py``) with
+    two additions:
+
+      1. The ``double_blocks`` loop runs entirely on cuda:0. Its outputs
+         (``img``, ``txt``) are concatenated (``cat(txt, img, dim=1)``)
+         and fed into the ``single_blocks`` loop, also starting on cuda:0.
+      2. At ``single_blocks[split_at]`` — the single PCIe boundary —
+         bridge to cuda:1 every tensor the downstream single blocks +
+         ``final_layer`` consume: ``img`` (the concatenated tensor),
+         ``pe``, ``txt_img_mask``, and a freshly-built ``mod_vectors_dict``
+         containing the cuda:1-resident slices for ``single_blocks[i].modulation.lin``
+         (``i >= split_at``) plus ``final_layer.adaLN_modulation.1``. The
+         original cuda:0 ``mod_vectors_dict`` is NOT mutated — gradient
+         checkpointing in the cuda:0 single blocks captures it by closure
+         and re-reads it during backward; mutating would corrupt that
+         re-read (Qwen-Image lesson).
+
+    After ``final_layer`` the output is moved back to cuda:0 so downstream
+    loss ops (which run on the model's nominal device) work without a
+    device-mismatch error.
+
+    Signature and return type match the vendored forward exactly.
+    """
+    import torch.utils.checkpoint as ckpt
+
+    from .src.layers import distribute_modulations, timestep_embedding
+
+    def forward(
+        self,
+        img: torch.Tensor,
+        img_ids: torch.Tensor,
+        txt: torch.Tensor,
+        txt_ids: torch.Tensor,
+        txt_mask: torch.Tensor,
+        timesteps: torch.Tensor,
+        guidance: torch.Tensor,
+        attn_padding: int = 1,
+    ) -> torch.Tensor:
+        if img.ndim != 3 or txt.ndim != 3:
+            raise ValueError("Input img and txt tensors must have 3 dimensions.")
+
+        # Devices of the first single block on each side of the boundary.
+        # double_blocks[*] + single_blocks[:split_at] all live on d0_dev;
+        # single_blocks[split_at:] + final_layer live on d1_dev.
+        d0_dev = next(self.single_blocks[0].parameters()).device
+        d1_dev = next(self.single_blocks[split_at].parameters()).device
+
+        # img_in / txt_in live on cuda:0; ensure inputs match.
+        if img.device != d0_dev:
+            img = img.to(d0_dev)
+        if txt.device != d0_dev:
+            txt = txt.to(d0_dev)
+        if img_ids.device != d0_dev:
+            img_ids = img_ids.to(d0_dev)
+        if txt_ids.device != d0_dev:
+            txt_ids = txt_ids.to(d0_dev)
+        if txt_mask.device != d0_dev:
+            txt_mask = txt_mask.to(d0_dev)
+
+        img = self.img_in(img)
+        txt = self.txt_in(txt)
+
+        # Distillation approximator — runs under no_grad on cuda:0,
+        # produces the mod_vectors tensor consumed by every block. The
+        # final tensor itself requires_grad_(True) per the vendored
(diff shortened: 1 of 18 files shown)


genno-whittlery:
Update: I've folded **Wan 2.2 dual-GPU support** into this PR (commit `010bb01`) and expanded the title/description accordingly.

Rationale for keeping it one PR rather than a separate one: both ports reuse the same toolkit-wide infrastructure — the `ToolkitNetworkMixin.force_to` / `LoRANetwork.apply_to` device-routing patches and the CPU-quantize-before-move load path. Separate PRs would conflict on those. The dual-GPU machinery isn't FLUX-specific; it's a per-trainer capability.

New in this update:
- `extensions_built_in/diffusion_models/wan22/wan22_dual_gpu.py` — `Wan22DualGPUMixin`, mirrors `Flux2DualGPUMixin`
- `Wan225bModel` / `Wan2214bModel` inherit it; `Wan21.load_wan_transformer` softens its "not supported" raise to dispatch to the mixin under `WAN_DUAL_GPU=true`
- 14B A14B handled per-expert (each of the two transformers distributed individually after quantize; the `DualWanTransformer3DModel` wrapper is untouched)

Validated on 2× RTX 5090: Wan 2.2 TI2V-5B at 1.86 it/s, T2V-A14B at 1.09 s/it (both experts exercised). Also added klein-4B/9B validation to the FLUX.2 side — the earlier description said klein was "not yet validated"; it is now (2.33 it/s / 1.21 it/s).

Happy to split Wan back out into its own PR if you'd rather review them separately — just say the word.

FurkanGozukara:
this is linux only or works on Windows as well?

genno-whittlery:
Update: **LTX-2** dual-GPU support is now folded in too (commit `c09e443`), title/description updated.

`extensions_built_in/diffusion_models/ltx2/ltx2_dual_gpu.py` — `LTX2DualGPUMixin`, mirrors the Wan mixin. `LTX2Model` inherits it. Single-boundary split forward: LTX-2 builds its modulation tensors once before the block loop, so the cuda:1 crossing at the split index bridges both modality streams (`hidden_states` + `audio_hidden_states`), both text streams, all precomputed modulation tensors, and all four computed RoPE tuples in one shot.

Four `ltx2.py` call sites that hardcoded `device_torch` are now routed to `te_device_torch` so the 12B Gemma3 text encoder lands on CPU under `LTX2_DUAL_GPU` (no-op when inactive — `te_device_torch == device_torch`): the TE load, the post-pipe "make sure" re-move, `get_prompt_embeds` (encodes on CPU; `cache_text_embeddings` runs it once at start), and `get_noise_prediction` — where the connectors + audio-latents placement now uses a `transformer_io_device` pinned to cuda:0, since diffusers' `Model.device` (first-parameter device) is ambiguous for a split model and was resolving to cuda:1, landing the connectors module on the wrong device.

Validated on 2× RTX 5090: LTX-2, `--fp8_base`, 48 blocks split 24/24, Gemma3 12B TE on CPU, 20 steps at 2.78 s/it, loss stable, checkpoint saved.

This PR now covers **FLUX.2 + Wan 2.2 + LTX-2** — all sharing the same toolkit-wide LoRA-routing infrastructure. Still happy to split per-trainer if you'd rather review them separately.

genno-whittlery:
Update: **Qwen-Image** dual-GPU support folded in (commit `96605dc`), title/description updated.

`extensions_built_in/diffusion_models/qwen_image/qwen_image_dual_gpu.py` — `QwenImageDualGPUMixin`, mirrors the LTX-2 / Wan mixins. 60 `transformer_blocks` split 30/30; single-boundary forward bridging both looped streams (`hidden_states` + `encoder_hidden_states`), `temb`, the RoPE tuple, the joint attention mask, and `modulate_index`. Qwen-Image is actually the cleanest of the four — no module-level `nn.Parameter`, and its `QwenEmbedRope` transfers freqs lazily, so neither the `scale_shift_table` nor the `.to()`-misses-buffer gotchas apply.

Two bugs worth noting from validation:
- The diffusers `QwenImageTransformer2DModel.forward` signature on the training box carries a deprecated `txt_seq_lens` param that an older diffusers lacked — the recreated split forward now matches the installed version exactly.
- `joint_attention_kwargs` is a dict captured *by reference* in each block's gradient-checkpoint closure. Mutating its `attention_mask` entry to cuda:1 at the boundary corrupted the cuda:0 blocks' backward recompute (they re-read the mutated cuda:1 mask → `attn_bias on cuda:1, others on cuda:0`). Fixed by building a separate cuda:1 dict at the boundary instead of mutating the shared one.

Validated on 2× RTX 5090: Qwen-Image, `--fp8_base`, 60 blocks split 30/30, Qwen2.5-VL TE on CPU, 20 steps at 1.56 s/it, loss stable, checkpoint saved.

This PR now covers **FLUX.2 + Wan 2.2 + LTX-2 + Qwen-Image** — all on the same shared toolkit-wide LoRA-routing infrastructure. As always, happy to split per-trainer if you'd rather review separately.

genno-whittlery:
@FurkanGozukara The dual-GPU mixin patterns themselves are platform-neutral — they're pure PyTorch device routing (per-module `.to("cuda:0")` / `.to("cuda:1")`, `te_device_torch` plumbing for the text encoder, a small `transformer_io_device` pin to disambiguate the I/O side after the split). No Linux-only calls — no NCCL P2P primitives, no `cuda-ipc` mailboxes, no `numactl`, no cgroups, no `posix_spawn`.

That said, every validation run on this PR was on 2× RTX 5090 on Linux (Ubuntu, the maverick training box). Windows should work mechanically — `WAN_DUAL_GPU=true` / `FLUX2_DUAL_GPU=true` / etc. are normal env vars, and `cuda:0` / `cuda:1` device strings behave identically — but I haven't smoke-tested on Windows myself. If you give it a try and hit anything that looks platform-specific, drop a trace here and I'll fix it.

genno-whittlery:
Status: **feature-complete, ready for review.**

Scope is now stable — all 8 architectures (FLUX.2 / Wan 2.2 / LTX-2 / Qwen-Image / Chroma / FLUX.1-Kontext / HiDream-I1 / Z-Image) are folded in, validated on 2× RTX 5090, and share the same toolkit-wide infrastructure (the `ToolkitNetworkMixin.force_to` / `LoRANetwork.apply_to` device-routing patches and the CPU-quantize-before-move load path). No further fold-ins planned.

Happy to split this into per-trainer PRs if that's easier to review — let me know which way you prefer.

wobba:
Would this work on e.g. one 5090 32gb + one 4060 16gb? Maybe not split on middle?

genno-whittlery:
Yes — there's an env-var escape hatch for exactly this. `FLUX2_DUAL_GPU_SPLIT_AT=N` overrides the default midpoint (`n_single // 2`) and lets you push more single_blocks onto the bigger card. For a 32 GB + 16 GB pair, the realistic recipe is:

- Set `CUDA_VISIBLE_DEVICES=<5090>,<4060>` so the 5090 is `cuda:0` (it carries all of `double_blocks` + img_in/txt_in/embed + the bulk of `single_blocks`, ~24 GB-class load).
- Pick a high `FLUX2_DUAL_GPU_SPLIT_AT` so only the tail of `single_blocks` + `final_layer` ride the 4060. With FLUX.2's 24 single_blocks, something like `FLUX2_DUAL_GPU_SPLIT_AT=20` puts 20 on the 5090 and 4 on the 4060 — start there and adjust.
- Keep `FLUX2_TE_DEVICE=cpu` (Mistral text encoder on CPU is already part of the recipe; that's what makes the 32 GB side fit at all).
- If the 4060 OOMs on activations, drop `--resolution` or batch first; tighter splits help a little but the activation crossing is one per fwd/bwd and not the dominant cost.

Two honest caveats: (1) I only validated on 2× RTX 5090 — asymmetric pairs aren't in the test matrix yet, so consider this "should work in principle." (2) PCIe boundary throughput will hurt step time more on a 5090/4060 split than on 2× 5090 because the 4060 is likely on narrower lanes. Training will run; expect a couple-x slowdown per step.

Would love a report if you try it.

PR #845, opened 2026-05-25
Fix Wan22 offline training

rmatif:
Wires Wan 2.2 training to use packaged local VAE, text encoder, and adapter resources so prod does not reach Hugging Face at runtime.


rmatif:
Closing; opened against the wrong base repository.

PR #849, opened 2026-05-29
Add 'Stop & Save Checkpoint' button

entmike:
## Summary

Adds a new **Save & Stop** button next to the existing Stop/Pause button in the Job Action Bar. When clicked, it sets both the `stop` and `save_now` flags on the job record so the training loop will save a checkpoint at the current step before terminating.

### Changes

- **New API endpoint**: `ui/src/app/api/jobs/[jobID]/save_and_stop/route.ts`
  Sets `stop` and `save_now` DB flags, sends SIGINT to the process, then resets both flags and marks status 'stopped'.

- **New client function**: `saveAndStopJob` in `ui/src/utils/jobs.ts`
  Calls the `/save_and_stop` endpoint.

- **UI**: `ui/src/components/JobActionBar.tsx`
  Adds a `Save` icon button next to the existing `Pause` (Stop) button. Shows a confirmation dialog before calling `saveAndStopJob`.

### How it works

1. User clicks **Stop & Save** button
2. Confirmation dialog appears: *"This will stop the job and save a checkpoint before terminating."*
3. On confirm: `saveAndStopJob` calls the `/save_and_stop` endpoint
4. The endpoint sets `stop=true` and `save_now=true` on the job
5. The training loop (via `maybe_save()`) sees `save_now` and calls `self.save(step_num)` before the stop watcher terminates the process
6. SIGINT is sent, DB flags are cleared, status set to 'stopped'

### Why this is useful

The existing **Stop** button sends SIGINT immediately, which can leave training mid-step with no checkpoint. This new button ensures a checkpoint is saved at the current step before termination, preventing loss of training progress.

### DIFF
--- a/ui/src/app/api/jobs/[jobID]/save_and_stop/route.ts
+++ b/ui/src/app/api/jobs/[jobID]/save_and_stop/route.ts
@@ -1,0 +1,56 @@
+import { NextRequest, NextResponse } from 'next/server';
+import { PrismaClient } from '@prisma/client';
+
+const prisma = new PrismaClient();
+
+export async function GET(request: NextRequest, { params }: { params: { jobID: string } }) {
+  const { jobID } = await params;
+
+  const job = await prisma.job.findUnique({
+    where: { id: jobID },
+  });
+
+  if (!job) {
+    return NextResponse.json({ error: 'Job not found' }, { status: 404 });
+  }
+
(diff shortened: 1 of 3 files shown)


entmike:
Ugh, sorry.  My agent got too excited with the PR.  I need to finish testing this first.

PR #863, opened 2026-06-05
Add support for remote dataset source plugins

lindkvis:
Introduce a plugin system that lets remote datasets be browsed and imported from external services directly in the UI.

* toolkit/dataset_sources/: base class, registry, and a CLI bridge that the Next.js API routes call to list plugins, browse items, fetch thumbnails, and build import job configs. Plugins live in extensions/ and self-register from their __init__.py, so they are only loaded on demand.
* ui: a RemoteSourceBrowseModal plus /api/datasets/remote/* routes for listing plugins, browsing, thumbnails, and importing, with the settings page driven by each plugin's settings_schema.

The framework is plugin-agnostic; no plugin is bundled here. There is a second PR coming that offers a PixlStash plugin (https://pixlstash.dev) but it should be possible to create other plugins (i.e. Immich).

### DIFF
--- a/toolkit/dataset_sources/__init__.py
+++ b/toolkit/dataset_sources/__init__.py
@@ -1,0 +1,12 @@
+from .base import RemoteDatasetSource, SettingField, SourceItem
+from .registry import register_source, get_source, get_all_sources, resolve_dataset_source
+
+__all__ = [
+    "RemoteDatasetSource",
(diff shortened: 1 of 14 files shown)


lindkvis:
I don't know if there is interest in this PR but I have resolved the conflicts anyway.

PR #866, opened 2026-06-07
Enhanced torch.compile System with Block-Level Compilation and Unified Whole-Model Fallback

PlagueKind:
# PR: Enhanced torch.compile System with Block-Level Compilation and Unified Whole-Model Fallback

This PR introduces a refined and more flexible torch.compile integration for AI Toolkit, improving both performance control and architectural adaptability across different model types.

---

## Overview

The compilation system is now primarily driven by a single required configuration flag:

* `compile: true`

This enables a dual-path compilation strategy depending on model structure and optional user configuration.

---

### Block-Level Compilation (`block_compile: true`)

When enabled, the system attempts to compile transformer blocks individually:
* `compile: true`
* `block_compile: true`

* Targets known transformer block containers across architectures
* Applies `torch.compile` per-block for finer granularity
* Skips modules protected by offloading hooks
* Automatically counts successfully compiled blocks

### Automatic Fallback Behavior

If no compatible blocks are found:

* The system automatically falls back to whole-model compilation
* Ensures compatibility across all model architectures
* Preserves stability when block-level compilation is not possible

---

## Whole-Model Compilation

When block-level compilation is disabled or falls back:

* The full UNet model is compiled using `torch.compile`
* Maintains consistent behavior across all model types
* Applies identical compile configuration parameters

---

## Quantized Model Handling

The system includes explicit safety checks for quantized models:

* Quantized models combined with transformer offloading are incompatible with compilation
* In this case, compilation is skipped with a warning
* This safety rule applies across both block-level and whole-model compilation paths

Where applicable:

* `fullgraph` is automatically adjusted to ensure compatibility during whole-model compilation

---

## Text Encoder Awareness

The system now detects whether the text encoder is offloaded:

* Prevents unnecessary device transfers for offloaded models
* Ensures compilation does not trigger unwanted memory movement
* Improves startup stability in memory-constrained setups

---

## Cache Configuration

The compiler cache size is now configurable:

* Default: `8`
* Configurable via `cache_size_limit`

Behavior:

* Custom values are applied directly to `torch._dynamo.config.cache_size_limit`
* Logging is only shown when a non-default value is used
* Keeps startup logs clean while still exposing overrides when relevant

---

### Core Options

* `compile: true` → enables compilation system
* `block_compile: true` → enables block-level transformer compilation

---

### Tuning Parameters

* `compile_mode`:

  * `default`
  * `reduce-overhead`
  * `max-autotune`
  * `max-autotune-no-cudagraphs`

* `compile_dynamic: true | false`

* `compile_fullgraph: true | false`

* `cache_size_limit: int` (default: 8)

---

## Logging Behavior

* Compilation mode is logged once per run in a consolidated format
* Cache size override is only printed when explicitly modified
* Block compilation reports number of successfully compiled transformer blocks
* Clear fallback messaging is shown when block compilation is not possible

---

## Design Notes

* Block-level compilation is preferred when enabled
* Whole-model compilation serves as a guaranteed fallback path
* All compilation paths share unified safety checks for quantization and offloading
* Text encoder awareness improves stability in memory-managed environments
* System prioritizes compatibility and graceful degradation over aggressive optimization

---

## Future Improvements

Future work will focus on:

* Expanding compatibility of block-level compilation across more transformer variants
* Reducing reliance on fallback whole-model compilation through improved model introspection
* Improving compile graph stability to allow quantized + offloaded configurations

Speed comparison of ideogram4 without compile vs with block compile


PlagueKind:
rewrote the code to streamline it, and retain the base function of normal compile as flag compile: true
with the added function block_compile: true as a secondary flag.
it's more robust, quicker startup as well as it doesn't sniff the text encoder offload.

Also removed unicode emoji to keep it compatible with windows without additional dependency.

PlagueKind:
Speed comparison of ideogram4 without compile vs with block compile

jaretburkett:
Thank you!

PR #868, opened 2026-06-07
Fix text encoder offload bug when caching embeddings

PlagueKind:
## Problem

When cached text embeddings are enabled, the text encoder may still be passed through memory management during training.

This creates unstable behavior where the text encoder is repeatedly attached and detached or transitions between managed states instead of being fully released from the active memory pipeline.

This results in unnecessary overhead and can cause inconsistent behavior during training, as well as increased VRAM usage.

## Root Cause

The text encoder lifecycle was not properly finalized when switching to cached embeddings, which allowed the layer offloading system to continue interacting with it during training.

In particular, the encoder was not fully detached from memory management after embeddings were cached, leaving it partially tracked.

## Fix

Ensure the text encoder is fully detached from memory management once cached embeddings are active.

This prevents repeated state transitions (ping-pong behavior) and removes it from the active offload pipeline.

## Testing

- Verified cached embedding workflow
- Verified text encoder is no longer managed after caching
- Verified no regression in standard (non-cached) training path

### DIFF
--- a/toolkit/memory_management/manager.py
+++ b/toolkit/memory_management/manager.py
@@ -1,5 +1,5 @@
 import torch
-from .manager_modules import LinearLayerMemoryManager, ConvLayerMemoryManager
+from .manager_modules import LinearLayerMemoryManager, ConvLayerMemoryManager, _DEVICE_STATE
 import random
 
 LINEAR_MODULES = [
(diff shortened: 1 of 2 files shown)


PlagueKind:
@jaretburkett Hopefully this is cleaner

jaretburkett:
Screenshot 2026-06-07 at 2 05 24 PM
Screenshot 2026-06-07 at 2 07 51 PM
Wow huge improvement. Thank you!

PlagueKind:
any time

PR #882, opened 2026-06-13
Make the final checkpoint follow the step numbering convention

fatalis:
Many people have requested this in Discord here https://discord.com/channels/1144033039786188891/1490993758496620564/1490993758496620564 . This makes the checkpoints consistent when resuming from a finished run after increasing max steps.


PlagueKind:
Needed

jaretburkett:
The problem is, a lot of companies rely on the save not having a step and this will break all of their systems. I know of 3 off the top of my head that would immediately break with this change.

PR #887, opened 2026-06-16
fix ideogram loading and lowvram fix

PlagueKind:
Ideogram 4 unquantized transformer as well as quantized text encoder were not behaving correctly with low vram.
cuda was also not accelerating their loading and  quantization.

This fixes that as well as significantly speeding up model loading.


PlagueKind:
Sync to main and removed extra lines

PR #899, opened 2026-06-19
Compile improvements - auto cache size, fix fullgraph setting, fix triton detection

PlagueKind:
## Summary
This PR builds on the previous torch compile improvements, with further refinements to the block compile function focused on stability, correctness, and startup performance.

### Automatic cache size limit detection

Adds autodetection of cache_size_limit based on the number of detected transformer blocks. This reduces unnecessary recompiles, improves compatibility, and speeds up execution. Manually specified cache_size_limit values are still respected, but if a user-supplied value is too low for the detected block count, it's automatically increased to prevent issues.

### Fullgraph setting fix

Corrects the fullgraph setting mismatch between config modules and the actual code, which was previously misconfigured and not always being applied or behaving as intended.

### Triton detection simplified to fail-fast

Triton no longer is being detected. Instead, it fails immediately and falls back to no compile with a print. This avoids a significant slowdown on Linux systems where the check was unnecessary, and results in a faster, cleaner fallback path overall for windows/linux users without Triton. Also added a function that kicks compiled blocks and falls back to training without compile when a mutation is detected mid compile.

### Impact

Improved overall stability
Startup speed restored to (or better than) the original baseline from my first PR 
Fewer recompiles and better compile-cache compatibility across varying model sizes
Possible iteration speed improvements

### DIFF
--- a/jobs/process/BaseSDTrainProcess.py
+++ b/jobs/process/BaseSDTrainProcess.py
@@ -2068,6 +2068,7 @@ class BaseSDTrainProcess(BaseTrainProcess):
         #     -> block-level compilation
         # ============================================================
         if self.model_config.compile:
+            compiled_refs = []  # (block_list, index, original_block) for rollback on failure
             try:
                 inner_unet_check = unwrap_model(self.sd.unet)
                 is_unet_offloaded = hasattr(inner_unet_check, '_memory_manager')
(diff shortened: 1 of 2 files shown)


PlagueKind:
Removed forced torchao with
[447773f](https://github.com/ostris/ai-toolkit/pull/899/commits/447773fb4a0fc119f4137694f5aeab64edb9f05f)
As it is no longer needed and modern versions of pytorch cause issues compared to Quanto.
Quantized unet offloading remains intact.

@jaretburkett I have also tested Klein4b the model with the issue you were having, and with the fixes of this whole PR I am unable replicate the issue, I believe at this point the forced Torchao is redundant and a hindrance as it only seems to cause severe startup slowdown.

fatalis:
7 minute first forward pass -> 1 minute on L40S 48GB w/ qwen image edit 2509 float8 quant

PlagueKind:
5060ti 16gb
Ideogram4 - Float8

Before PR
First forward pass = 7-12 minutes
After PR
First forward pass = barely slower than with compile off

PlagueKind:
Still yet to find a conflict @jaretburkett

PR #900, opened 2026-06-19
Fix jobactionbar spacing web and mobile

PlagueKind:
Alignment spacing for the jobactionbar buttons on the web ui were incorrect and off balanced. (More space between cogwheel and other buttons)
I've fixed that as well as slightly padding those buttons correctly for mobile users so they can comfortably be clicked without accidentally hitting the wrong button.


PlagueKind:
All buttons align correctly on different modes and pages now.

PR #901, opened 2026-06-19
Fix progress bar iteration speed accuracy and speed

PlagueKind:
This change replaces tqdm’s internal rate estimation with a deterministic, time-based EMA smoothing approach.

Problem:
- tqdm’s built-in rate estimator becomes unstable under non-uniform step timing
- save/sample/log pauses cause delayed convergence and “catch-up” behavior in reported it/s

Solution:
- Introduce external time-based rate calculation using perf_counter()
- Apply lightweight EMA smoothing to stabilize display without lag
- Preserve existing postfix metrics system (losses, lr, etc.)
- Avoid modifying training loop behavior, only affects progress display

Result:
- Stable iteration speed reporting
- No catch-up artifacts after pauses
- No dependency on tqdm internal smoothing model

### DIFF
--- a/toolkit/progress_bar.py
+++ b/toolkit/progress_bar.py
@@ -1,9 +1,12 @@
 from tqdm import tqdm
-import time
 
 
 class ToolkitProgressBar(tqdm):
     def __init__(self, *args, **kwargs):
+        kwargs.setdefault("mininterval", 1.0)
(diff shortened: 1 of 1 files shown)


PlagueKind:
Switched to simplified code by @fatalis

jaretburkett:
Fixed with 27a03a9

PR #909, opened 2026-06-24
Force HF downloads to show progress bars when output is not a TTY

fatalis:
No progress bars on downloads have been causing a lot of headaches for users, so here's this monkey patch that fixes it. There's no other good way to do this without changing every hf_download call in the repo.


jaretburkett:
Thank you

PR #912, opened 2026-06-24
Fix most typos in the repo

fatalis:
Agent fixed typos in the repo and I manually reviewed the diff.


MaxKruse:
i have a suspicion that some of these typos were meant to be "vell" -> "bell". not "well".

fatalis:
> i have a suspicion that some of these typos were meant to be "vell" -> "bell". not "well".

good catch, will fix soon

PR #915, opened 2026-06-26
Fix Krea 2 README categorization from Video to Image

alpaim:
Krea 2 was listed under Video models in the README, but it's a text-to-image model only. Moved it to the Image section where it belongs.


DaveGravel:
Same in this section:
image

jaretburkett:
Fixed with 6798475

PR #916, opened 2026-06-27
Fix ZIT sampling defaults to be more inline with reference implementation

fatalis:
CFG `0` is the off-switch for ZIT
`9` steps makes 8 useful steps and match sigmas of the ZIT ref impl.


fatalis:
fixed in https://github.com/ostris/ai-toolkit/commit/6233efe1bbefd243caf1d30f8354ca275b3980f6

Issue #908, opened 2026-06-24
Repeated job initialisation and shutdown after quickly stopping and starting a job queue.


mcDandy:
## This is for bugs only

Did you already ask [in the discord](https://discord.gg/VXmU2f5WEU)?

Yes/No

You verified that this is a bug and not a feature request or question by asking [in the discord](https://discord.gg/VXmU2f5WEU)?

Yes/No

## Describe the bug
A job is repeatly trying to start before stopping when I stop queue with running job and imidietly start it again. After stopping the job and later starting it again, it works as expacted

Console output is
```
[WORKER] Starting job fa739e35-46ea-46c9-979c-d70f0efe6fe2 on GPU(s) 0
[WORKER] Error moving log file: Error: EBUSY: resource busy or locked, rename 'D:\stability\Data\Packages\AI-Toolkit\output\plfun4-krea2_lokr\log.txt' -> 'D:\stability\Data\Packages\AI-Toolkit\output\plfun4-krea2_lokr\logs\4_log.txt'
[WORKER]     at Object.renameSync (node:fs:1033:11)
[WORKER]     at D:\stability\Data\Packages\AI-Toolkit\ui\dist\cron\actions\startJob.js:40:30 {
[WORKER]   errno: -4082,
[WORKER]   code: 'EBUSY',
[WORKER]   syscall: 'rename',
[WORKER]   path: 'D:\\stability\\Data\\Packages\\AI-Toolkit\\output\\plfun4-hidream-o1_64_copy_copy_copy\\log.txt',
[WORKER]   dest: 'D:\\stability\\Data\\Packages\\AI-Toolkit\\output\\plfun4-hidream-o1_64_copy_copy_copy\\logs\\4_log.txt'
[WORKER] }
```

output in browser
```
...
Using SQLite database at D:\stability\Data\Packages\AI-Toolkit\aitk_db.db
Job ID: "fa739e35-46ea-46c9-979c-d70f0efe6fe2"
#############################################
# Running job: plfun4-krea2_lokr
#############################################
Running  1 process
Error running job: Job returning to queue
========================================
Result:
 - 0 completed jobs
 - 1 failure
========================================
Traceback (most recent call last):
Traceback (most recent call last):
  File "D:\stability\Data\Packages\AI-Toolkit\run.py", line 135, in <module>
  File "D:\stability\Data\Packages\AI-Toolkit\run.py", line 135, in <module>
        main()main()
[...]

PR #920, opened 2026-06-28
Clear stale return-to-queue flag when starting jobs, fixes crash loop

fatalis:
Fixes a notorious crash loop some people have been resorting to editing the sqlite database manually to fix.


fatalis:
Fixes #908

PR #922, opened 2026-06-29
Fix several cases of silent crashing on UI

fatalis:
* The UI now opens the log FD and passes it to the child process (stays alive after UI exit)
* Use PYTHONUNBUFFERED to possibly log errors before python exits
* Errors with imports / python are now logged
* Fix DB not stopping job on some types of crashes


jaretburkett:
Thank you

PR #923, opened 2026-06-29
Relocate Krea 2 to Images section

astreloff:
## Description

Moved Krea 2 out of Video and into the Images tab and cleaned up some whitespace.


jaretburkett:
Fixed with 6798475

PR #927, opened 2026-07-02
Fix CUDA OOM during Krea2 block-by-block quantization

thedivergentai:
## Summary
- Fix CUDA OOM during Krea2 transformer quantization by synchronizing and flushing GPU memory after each block is offloaded to CPU.
- Ensure Krea2 starts quantization from a clean CPU baseline, including assistant LoRA weights when present.

## Test plan
- [x] Reproduced original failure path (OOM around block 25/28 during `quantize_model`)
- [x] Verified all 28 blocks quantize successfully on a 24GB RTX 3090
- [x] Confirmed Krea2 Raw training job completes model load and starts training


thedivergentai:
Closing: upstream already fixed the same root cause in 933ca1c (blocking CPU offload during block quantization / pin memory leak).

Issue #374, opened 2025-08-15
A few issues with 'cache_text_embeddings'


GreenLandisaLie:
I tried this feature with Kontext (4bit) using the latest version of AI-Toolkit plus a manual merge of this PR: [https://github.com/ostris/ai-toolkit/pull/331]( https://github.com/ostris/ai-toolkit/pull/331) and I found a few issues that I believe are likely not related to that PR so I'm posting them here. Note that I've used that PR without 'cache_text_embeddings' and had no problems.

- toolkit\train_tools.py:
   - For some reason my text_encoder[0] (clip) was on cuda:0 but the text_encoder[1] (T5) was on cpu. This caused a crash in line: `prompt_embeds = text_encoder[1](text_input_ids.to(device), output_hidden_states=False)[0]`. I solved it by simply adding a check on both and sending them to cuda when not there - but if this is something not caused by the PR then you would need a more robust solution.

- extensions_built_in\diffusion_models\flux_kontext\flux_kontext.py:
   - The get_te_has_grad(self) function was being called after TE had been turned into a FakeTextEncoder. This caused a crash right away because that function tries to access the 'encoder' property which does not exist in FakeTextEncoder. I solved it by simply changing it into a try except block that returns False when it fails but again - if this is not being caused by the PR then you would need a better approach that solves this on all text encoders.

- toolkit\unloader.py:
   - The unload_text_encoder(model: "BaseModel") function is not clearing the allocated memory. Simply replacing the text encoders with a FakeTextEncoder will preserve its allocated space in memory. I made a shitty fix for myself by force sending all of the text encoders to cpu before replacing them with the FakeTextEncoder - because when you do so, torch will ""properly"" clear the allocated memory. Doing so solved my issue with the transformer not fitting completely in my 16Gb VRAM after 'unloading the TE'.

Another thing I noticed was that for some reason every time I restarted the same job - a few more TE cache embeds would be added into their cache folder. Instead of recognizing that all of them are already there - some keep being generated every time the job starts and they are not replacing the existing ones - instead they just keep pilling up on the cache folder.

t3t3:
And there is at least one more problem with that (at least how it happens in my case, when training Chroma):

[...]

PR #935, opened 2026-07-06
Send T5 tokens to the T5 encoder's own device when caching text embeddings

rajakshay:
## What

`encode_prompts_flux` in `toolkit/train_tools.py` computes a single `device` from the first text encoder (CLIP) and reuses it for the second encoder (T5). When caching text embeddings, it sends the T5 input ids to `device` (CLIP's device) instead of to T5's own device.

## Why it breaks

On setups that place the large T5 encoder on a different device than CLIP (e.g. CPU offload / `device_map` on low-VRAM or unified-memory systems), the tokens land on CLIP's GPU while T5's weights are elsewhere, so the T5 forward raises:

```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu
```

This surfaces during "Caching text embeddings to disk" — the model, VAE, network, and CLIP all load fine first, then it dies when T5 is invoked.

## Fix

Send the tokens to `text_encoder[1].device` (T5's own device).

## Regression safety

When both encoders are co-located (the default single-GPU case), `text_encoder[1].device == device`, so this is a **no-op** — no behavior change for existing setups. It only corrects the split-device path.

## Validation

- **Before:** a CLIP+T5 LoRA training run with the T5 encoder offloaded crashes at the caching step with the device mismatch above.
- **After:** caching completes and training proceeds; a full LoRA training run finished end-to-end and saved weights.

Refs #374 (the `cache_text_embeddings` device-mismatch report).

### DIFF
--- a/toolkit/train_tools.py
+++ b/toolkit/train_tools.py
@@ -561,7 +561,7 @@ def encode_prompts_flux(
     )
     text_input_ids = text_inputs.input_ids
 
-    prompt_embeds = text_encoder[1](text_input_ids.to(device), output_hidden_states=False)[0]
+    prompt_embeds = text_encoder[1](text_input_ids.to(text_encoder[1].device), output_hidden_states=False)[0]
 
     dtype = text_encoder[1].dtype
     prompt_embeds = prompt_embeds.to(dtype=dtype, device=device)


rajakshay:
### Reproduction (no models or data)

Tiny stand-in encoders — CLIP on GPU, T5 on CPU — which is exactly the state when the large T5 encoder is offloaded while CLIP stays on the GPU. Needs one CUDA device (to place the two encoders on different devices). Run from the repo root.

<details><summary>repro.py</summary>

```python
import sys
import torch
import torch.nn as nn

if not torch.cuda.is_available():
    print("SKIP: needs a CUDA device to place CLIP and T5 on different devices")
    sys.exit(0)

from toolkit.train_tools import encode_prompts_flux


class Tok:
    model_max_length = 77
    def __call__(self, prompts, **kw):
        o = type("O", (), {})()
        o.input_ids = torch.zeros((len(prompts), 8), dtype=torch.long)
        return o


class CLIP(nn.Module):
    def __init__(self):
        super().__init__()
        self.lin = nn.Linear(8, 8)
    @property
    def device(self): return next(self.parameters()).device
    @property
    def dtype(self): return next(self.parameters()).dtype
    def forward(self, ids, output_hidden_states=False):
        o = type("O", (), {})()
        o.pooler_output = self.lin(ids.float())
        return o


class T5(nn.Module):
    def __init__(self):
        super().__init__()
        self.emb = nn.Embedding(100, 16)
    @property
    def device(self): return next(self.parameters()).device
    @property
    def dtype(self): return next(self.parameters()).dtype
    def forward(self, ids, output_hidden_states=False):
        return (self.emb(ids),)


clip = CLIP().to("cuda")   # CLIP on GPU
t5 = T5().to("cpu")        # T5 offloaded to CPU (the real low-VRAM / device_map case)

try:
    embeds, pooled = encode_prompts_flux([Tok(), Tok()], [clip, t5], ["a photo of a cat"])
    print("PASSED: encode_prompts_flux returned embeds", tuple(embeds.shape),
          "pooled", tuple(pooled.shape))
except RuntimeError as e:
    print("REPRODUCED RuntimeError:", str(e).splitlines()[0])
```
</details>

Result:

```
# BEFORE (current main)
REPRODUCED RuntimeError: Expected all tensors to be on the same device, but got index is on cuda:0, different from other tensors on cpu (when checking argument in method wrapper_CUDA__index_select)

# AFTER (this PR)
PASSED: encode_prompts_flux returned embeds (1, 8, 16) pooled (1, 8)
```

rajakshay:
Hi @jaretburkett 
I'd appreciate it if you can point me in the right direction to find a reviewer for this pull request. Thanks!

Issue #416, opened 2025-09-05
train_config expected during inference


mbesenzoni:
Training and generating with FLUX schnell locally I've encountered the following error while trying to generate images, after the LORA is trained:

``` python
Traceback (most recent call last):
  File "/app/ai-toolkit/finetune.py", line 332, in run_job
    job = get_job(config_file, job_name)
  File "/app/ai-toolkit/toolkit/job.py", line 26, in get_job
    return GenerateJob(config)
  File "/app/ai-toolkit/jobs/GenerateJob.py", line 16, in __init__
    self.load_processes(process_dict)
  File "/app/ai-toolkit/jobs/BaseJob.py", line 64, in load_processes
    self.process.append(ProcessClass(i, self, process))
  File "/app/ai-toolkit/jobs/process/GenerateProcess.py", line 105, in __init__
    self.train_config.noise_scheduler,
AttributeError: 'GenerateProcess' object has no attribute 'train_config'
```

So I investigated the `GenerateProcess` class and at line 105 a `train_config` is expected to retrieve the `noise_scheduler`:

https://github.com/ostris/ai-toolkit/blob/645046701b9816e3ffb2b3931aeef3b0171680cc/jobs/process/GenerateProcess.py#L105

This seems weird to me as I'm trying to run inference and the `yaml`config file, based on the examples available on github + whatever I could find online, only has a `generate` section with the expected `noise_scheduler` (in my case `flowmarch`):

https://github.com/ostris/ai-toolkit/blob/645046701b9816e3ffb2b3931aeef3b0171680cc/config/examples/generate.example.yaml#L13

Is this a bug or am I supposed to have a `train` section in an inference config file?

PR #936, opened 2026-07-06
Use the generate config sampler instead of the undefined train_config in GenerateProcess

rajakshay:
## What

In `GenerateProcess.__init__` (`jobs/process/GenerateProcess.py`), the sampler-building branch references `self.train_config.noise_scheduler`, but `GenerateProcess` never sets `self.train_config` (only training processes do).

## Why it breaks

Running a `generate` job that reaches this branch crashes at construction:

```
AttributeError: 'GenerateProcess' object has no attribute 'train_config'
```

## Fix

Use `self.generate_config.sampler`, which is already built earlier in the same `__init__`. `GenerateConfig` defines `sampler` (default `'ddpm'`), which is exactly what `get_sampler(...)` expects.

## Note on the previous attempt (#498)

This revives the fix reported in #416 and attempted in #498 (closed for inactivity). It intentionally uses `generate_config.sampler` rather than `generate_config.noise_scheduler` (as in #498): `GenerateConfig` has a `sampler` attribute but **no** `noise_scheduler` attribute, so `.sampler` actually resolves the crash instead of moving it to a different `AttributeError`.

## Regression safety

This branch currently always crashes, so no working path is affected.

## Validation

- **Before:** a FLUX `generate` job crashes at init with the `AttributeError` above.
- **After:** the generate job constructs, loads the model, and produces images (validated with a FLUX generate run that produced a full batch).

Refs #416. Supersedes the approach in #498.

### DIFF
--- a/jobs/process/GenerateProcess.py
+++ b/jobs/process/GenerateProcess.py
@@ -102,7 +102,7 @@ class GenerateProcess(BaseProcess):
             if self.model_config.is_lumina2:
                 arch = 'lumina2'
             sampler = get_sampler(
-                self.train_config.noise_scheduler,
+                self.generate_config.sampler,
                 {
                     "prediction_type": "v_prediction" if self.model_config.is_v_pred else "epsilon",
                 },


rajakshay:
### Reproduction (no models or data, CPU-only)

`get_model_class` is stubbed so no weights load and so the model class has no `get_train_scheduler` (which forces the sampler else-branch containing the bug). The crash happens before any model would be loaded. Run from the repo root.

<details><summary>repro.py</summary>

```python
import importlib
from collections import OrderedDict

# import_module returns the real submodule even though jobs.process re-exports the class
GP = importlib.import_module("jobs.process.GenerateProcess")


class DummyModel:  # no get_train_scheduler -> forces the sampler else-branch
    def __init__(self, *a, **k):
        pass


GP.get_model_class = lambda cfg: DummyModel  # no weights loaded


class FakeJob:
    name = "repro"
    meta = OrderedDict()
    device = "cpu"


config = OrderedDict({
    "output_folder": "/tmp/repro_out",
    "model": OrderedDict({"name_or_path": "stub", "is_flux": True}),
    "generate": OrderedDict({"prompts": ["a photo"], "sampler": "ddpm"}),
    "dtype": "float16",
})

try:
    GP.GenerateProcess(0, FakeJob(), config)
    print("PASSED: GenerateProcess constructed; sampler built from generate_config.sampler")
except AttributeError as e:
    print("REPRODUCED AttributeError:", str(e))
except Exception as e:
    print("OTHER:", type(e).__name__, "-", str(e).splitlines()[0])
```
</details>

Result:

```
# BEFORE (current main)
REPRODUCED AttributeError: 'GenerateProcess' object has no attribute 'train_config'

# AFTER (this PR)
PASSED: GenerateProcess constructed; sampler built from generate_config.sampler
```

PR #945, opened 2026-07-14
LoKr forward pass optimization + convrot compile fix

PlagueKind:
### What this does
LoKr's forward pass was rebuilding the full adapter weight matrix from scratch every single training step, even though the adapter itself only needs a small fraction of that size to store. This patch skips that step and computes the same result directly from the small pieces instead — same math, much less work per step.

### Why it matters
- Faster training steps, especially backward pass
- Very slight lower VRAM usage
- Fixes a `torch.compile` slowdown specific to convrot-quantized (OstrisLinear) layers — those layers were breaking compile's ability to fuse operations efficiently, which made training run close to compile-disabled speed. This patch resolves that so compile actually helps again on those layers.

### What's unchanged
- Training precision/quality — same dtype compute path as before, just less redundant work
- Conv2d-based LoKr layers — untouched, still using the original code path
- No config changes needed, works automatically

### DIFF
--- a/toolkit/models/lokr.py
+++ b/toolkit/models/lokr.py
@@ -103,6 +103,9 @@ class LokrModule(ToolkitModuleMixin, nn.Module):
         self.use_w1 = False
         self.use_w2 = False
         self.can_merge_in = True
+        # only the plain Linear path (no cp/Conv2d) gets the factorized forward;
+        # Conv2d and the cp branch keep using get_weight()/make_kron() below.
+        self._fast_linear = False
 
         # avoid the weight property on quantized OstrisLinear: it dequantizes the
         # whole weight just to answer .shape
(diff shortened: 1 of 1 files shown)


fatalis:
5080 @ 512 Krea2-Raw

compile off:
3s/it -> 1s/it

compile on:
1.9s/it -> 1s/it

PlagueKind:
> 5080 @ 512 Krea2-Raw
> 
> compile off: 3s/it -> 1s/it
> 
> compile on: 1.9s/it -> 1s/it

Seems like larger memory bus cards benefit even more 🔥

jaretburkett:
Great work. Seeing a large speed increase on my test as well.

PlagueKind:
> Great work. Seeing a large speed increase on my test as well.

Thanks!

PR #946, opened 2026-07-14
Fix adapter scalar handling under torch.compile

Zironic:
This change is indended to prevent loras and their derivatives from being compiled as CPU scalars instead of CUDA scalars.


jaretburkett:
Looks good. I added one check for one adapter merging issue. Thank you

PR #947, opened 2026-07-14
krea2: fix timestep-shift training

tcldr:
## Description

With `timestep_type: shift`, krea2 trains on a far more high-noise-biased timestep distribution than intended because `BaseSDTrainProcess` is expecting to find the token patch size via `patch_size` rather than the `patch` config key used by the Krea 2 types – so the lookup falls back to the incorrect `patch_size = 1`

## Fix

Probably a few ways to do this, but figured renaming the config type from `patch` to `patch_size` – so that BaseSDTrainProcess can find the expected property – and updating all internal consumers would be cleanest.

### DIFF
--- a/extensions_built_in/diffusion_models/krea2/krea2.py
+++ b/extensions_built_in/diffusion_models/krea2/krea2.py
@@ -69,7 +69,7 @@ KREA2_MMDIT_CONFIG = dict(
     kvheads=12,
     multiplier=4,
     layers=28,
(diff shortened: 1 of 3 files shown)


dremme:
Is this still relevant? I'm training a lot of Krea 2 stuff with shift, but haven't noticed issues?

jaretburkett:
Fixed with e65c4d0

PR #949, opened 2026-07-14
Add sample_start_step setting to configure when sampling starts

fatalis:
Requested feature in the Discord.

* Sets the mininum step number when sampling will run
* Baseline samples will still run unless "Skip First Sample" is enabled


jaretburkett:
Thank you

PR #957, opened 2026-07-17
Honor model-declared patch size

Zironic:
This fixes the patch name incompatibility for Krea2 and any other model with a similar issue.


jaretburkett:
Fixed with e65c4d0

PR #960, opened 2026-07-17
captioner: fix Ideogram4 prompt Unicode escape syntax

Zironic:
## Description

Current code declares the prompt as a normal string while containing the literal \uNNNN. Python interprets that as a malformed Unicode escape, so the module can fail during import.

Fix is to just add the raw-string prefix.


jaretburkett:
Fixed in 85a6880

PR #961, opened 2026-07-17
prompt cache: preserve frozen tensor dtypes across serialization

Zironic:
## Description

The way that frozen_dtype_keys are currently handled is lost during save() and load(). This code adds proper serialization.


jaretburkett:
Fixed in f421542

PR #979, opened 2026-07-26
add per image adaptive lr with logging

PlagueKind:
**Title:** Add per-image adaptive LR (opt-in)

**Description:**

Tracks each dataset image's loss trend over a step-count window (~one dataset pass, auto-computed from `dataset_size / batch_size`) and throttles images that stay hard without improving — likely a bad/mismatched caption — while giving a small boost to consistently healthy ones. Escalates the throttle the longer an image stays stuck.

Model- and network-agnostic: hooks into `SDTrainer.calculate_loss`, the single shared loss path for every architecture and both LoKr and LoRA, and writes into the existing per-item `loss_multiplier` field rather than adding a new mechanism.

**Changes**
- `toolkit/loss_watch.py` (new) — `PerImageAdaptiveLR` watcher: timestep-bucket-normalized residuals, vote-persistence classification (stuck/suspect/exhausted/healthy), escalating throttle multiplier. With logging
- `toolkit/config_modules.py` — `per_image_adaptive_lr` toggle + `per_image_adaptive_lr_window_steps` override, both opt-in/default off.
- `extensions_built_in/sd_trainer/SDTrainer.py` — observes per-sample loss/timestep pre-multiplier in `calculate_loss`.
- `jobs/process/BaseSDTrainProcess.py` — instantiates watcher, step-count window boundary trigger, checkpoint save/load persistence (sidecar JSON, same pattern as `learnable_snr.json`).
- `ui/src/app/jobs/new/SimpleJob.tsx` + `docs.tsx` — UI toggle in the Training card, doc tooltip.

**Not in scope here:** auto-recaption of stuck images (needs a vision-capable captioner per-arch) — separate follow-up.

### DIFF
--- a/extensions_built_in/sd_trainer/SDTrainer.py
+++ b/extensions_built_in/sd_trainer/SDTrainer.py
@@ -920,11 +920,17 @@ class SDTrainer(BaseSDTrainProcess):
                 # (flow-matching schedulers included) — normalize to [0, 1] for the bucket index.
                 num_train_timesteps = float(getattr(self.train_config, 'num_train_timesteps', 1000) or 1000)
                 for idx, file_item in enumerate(batch.file_items):
+                    # nominal configured resolution (e.g. 256/512/1024), NOT the post-aspect-crop
+                    # dimensions — with bucketing, crop_width/crop_height vary per aspect ratio
+                    # within the same resolution tier (256/288/336/... are all "256"), which would
+                    # fragment one tier into several and defeat the point of grouping by it.
+                    res = int(getattr(getattr(file_item, 'dataset_config', None), 'resolution', 0) or 0)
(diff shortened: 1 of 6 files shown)


PlagueKind:
Made it resolution aware so one single resolution doesn't heavily skew the voting on an image.

mcDandy:
I hope it gets added. Helps a lot when you are training multi-concept and one of the sub-concepts Is hard for the model. Other sub-concepts do not get overdone.

PlagueKind:
> I hope it gets added. Helps a lot when you are training multi-concept and one of the sub-concepts Is hard for the model. Other sub-concepts do not get overdone.

Definitely helped learning a few difficult concepts for me. Also helps reduce overtraining

PR #999, opened 2026-08-07
Free the text encoder on unload instead of parking it on CPU

matlowai:
`unload_text_encoder` swaps in `FakeTextEncoder` but `_detach_and_cpu` moves the real module to CPU, where it stays resident for the rest of the run — stray references keep it alive through gc. With the MiniMax H3 Qwen3-VL 32B nvfp4 encoder that is **15.7 GB of host RAM** doing nothing (verified with a live-tensor census after unload: 13.7 GB uint8 + 1.2 GB bf16 + 0.8 GB int8, all text-encoder weights).

This changes the helper to move storages to the **meta device** — freeing them regardless of stray references — and adds a guarded `malloc_trim(0)`, since glibc otherwise retains ~3.4 GB of freed arenas (measured: with zero live tensors, RssAnon stayed at 4.2 GB until trim; 0.8 GB after).

Measured on an H3 LoRA run with `cache_text_embeddings: true`: ready-to-train host RssAnon **15,874 MB → 791 MB**. Trained a 600-step run plus sampling on the patched path with no regressions.

Semantics note: after the fake swap every public handle points at the fake, so any code running the real encoder post-unload was already unsupported — this makes the free explicit instead of accidental-retention.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/toolkit/memory_management/manager.py
+++ b/toolkit/memory_management/manager.py
@@ -165,11 +165,16 @@ class MemoryManager:
                     continue
 
     @classmethod
-    def detach(cls, module: torch.nn.Module):
+    def detach(cls, module: torch.nn.Module, unpin_cpu_tensors: bool = False):
         """
         Reverse of attach(). Moves unmanaged modules back to CPU, restores the
-        original .to() and forward methods on all child layers, unpins CPU weight
-        tensors, and clears the global CUDA device state.
+        original .to() and forward methods on all child layers, and clears the
+        global CUDA device state.
+
+        CPU weights are left pinned by default. Unpinning requires cloning each
+        pinned tensor, which temporarily doubles its host-memory allocation. Set
+        ``unpin_cpu_tensors=True`` only when the detached module will remain
+        resident and must release its pinned-memory allocations.
 
         Call this before unloading/replacing a module that had attach() applied.
         """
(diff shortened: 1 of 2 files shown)


matlowai:
Context on the motivating setup, and one design question:

Found on a box with 2× RTX 6000 Pro (96 GB VRAM each) but ~96 GB of host RAM — training MiniMax H3 there leaves huge VRAM headroom while the CPU-parked encoder plus a second concurrent trainer pushed host RAM into OOM-killing desktop apps. On smaller-RAM machines (32–64 GB) training H3-class models, the parked 15.7 GB is likely the difference between fitting and swapping.

Design question for you: happy to gate this behind a flag (e.g. `free_text_encoder_on_unload: true` by default) if there are flows that expect to revive the real encoder after `unload_text_encoder` — mid-run sample-prompt edits, etc. As far as I could trace, this path only runs when all embeddings (including sample prompts) are already cached and the FakeTextEncoder swap has made the real module unreachable through public handles, so freeing looked strictly better — but you know the roadmap. Trained a full run + sampling on the patched path with no regressions either way.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

matlowai:
Pushed f24d58f based on downstream feedback from Fatalis: `detach()` unpinned CPU weights by cloning them, so detach-then-free transiently doubled host memory for the module (enough to hit swap on tight boxes). Unpinning is now opt-in via `unpin_cpu_tensors=True` — the detach-then-free path (which this PR fixes) never needs it, since pinned allocations are released with the tensors. Credit to Fatalis for the report and patch.

fatalis:
Observed this behavior too, tested this and works for me. +1

jaretburkett:
Fixed with 682b27c

PR #1000, opened 2026-08-07
Add torchaudio to requirements

matlowai:
`toolkit/config_modules.py:7` imports `torchaudio` at module level (also used in `dataloader_mixins` and `toolkit/audio`), but it isn't declared in requirements, so a fresh `pip install -r requirements.txt` env fails on the first toolkit import.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sinadavinc-hash:
Sina

sinadavinc-hash (commented):
Siba

PR #1018, opened 2026-08-25
Fix sample image grid ordering for 10+ prompts

### LINKED ISSUE #1008: Image grid behaving strangly for videos.
## This is for bugs only

Did you already ask [in the discord](https://discord.gg/VXmU2f5WEU)?

Yes/No

You verified that this is a bug and not a feature request or question by asking [in the discord](https://discord.gg/VXmU2f5WEU)?

Yes/No

## Describe the bug

When I train minimax H3 generation grid is wierd. I have 14 prompts. With one using reference image (ID 13 with ctrl_img_1 and ctrl_img (since I copied that from different job)). I tried multiple browsers, updating AI-Toolkit to latest comit, cleaning cache using dev tools...

1. during generation only some can be acessed using arrows when visiting first image. It is untill the one with the last ID is hit. The last one is not however last in position.
2. The  samples are not ordered. Most times it is 1, 2, 11, 12, 13, 14 (gets stuck here, you must esc and click next one manually), 3, 4, 5, 6, 7, 8, 9, 10 and it is continued by 1 from the next set. That everything is from pressing right
3. Sometimes it is 1...9,11...14,10. Happened once. Also once it kept in order
4. Currently generated are ~~allways~~ for the first 9 images in order. They stop being in order when switching to different card and back once fhey are finished generating.
5. I ammost of the time on overview tab.

rossnot:
Closes #1008

Fixes the image grid scrambling (1, 2, 11, 12, 13, 14, 3, 4...) reported when training with 10 or more prompts.
  
`save_image_atomic` was called without `max_count` in `base_model.py` and `stable_diffusion_model.py`, so the `[count]` placeholder was never zero-padded. With 10+ prompts this produced unpadded filenames (0, 1, 10, 11...) which sort lexicographically in the wrong order. The `max_count` parameter already existed in `GenerateImageConfig` for exactly this purpose it just wasn't being passed at the call sites.
  
Also switches the samples API route sort to `localeCompare` with `numeric: true` as a safety net for existing outputs that were saved without zero-padding.

### DIFF
--- a/toolkit/models/base_model.py
+++ b/toolkit/models/base_model.py
@@ -708,7 +708,7 @@ class BaseModel:
                         extra,
                     )
 
-                    gen_config.save_image_atomic(img, i)
+                    gen_config.save_image_atomic(img, i, len(image_configs))
                     gen_config.log_image(img, i)
                     self._after_sample_image(i, len(image_configs))
                     flush()
--- a/toolkit/stable_diffusion_model.py
+++ b/toolkit/stable_diffusion_model.py
@@ -1728,7 +1728,7 @@ class StableDiffusion:
                             generator=generator,
                         ).images[0]
 
-                    gen_config.save_image_atomic(img, i)
+                    gen_config.save_image_atomic(img, i, len(image_configs))
                     gen_config.log_image(img, i)
                     self._after_sample_image(i, len(image_configs))
                     flush()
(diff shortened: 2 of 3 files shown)


mcDandy:
wierd thing is that for images it works correctly since the timestamp (linux milis) are correct for images, (probably the same as creation time of the file). For videos it is the timestamp of aither first video or when generation started (for every video, the total difference of the timestamp between them is < 1 ms).

rossnot:
Yeah, that tracks. Videos sharing a timestamp is why the zero-padding matters there.

PR #1023, opened 2026-08-29
Skip sampling instead of crashing when cache_text_embeddings faked the text encoder

hitoshic1982:
## Bug

With `cache_text_embeddings: true`, the text encoder is replaced by `FakeTextEncoder` (toolkit/unloader.py) to free memory. When a scheduled mid-training sample fires, prompt encoding hits the fake encoder and raises:

```
AttributeError: 'FakeTextEncoder' object has no attribute 'encoder'
```

which propagates out of `job.run()` and **kills the entire training run at the first sample step**. Observed on a Chroma LoRA run (Windows, RTX 5060 Ti 16GB): the process trained fine for ~80 minutes and died exactly at step 250, the first `sample_every` boundary. The step-250 checkpoint survived only because saving runs before sampling.

## Fix

Before sampling, detect whether the (possibly list-typed) text encoder is a `FakeTextEncoder`; if so, print an actionable warning — pointing at `disable_sampling: true`, or turning off `cache_text_embeddings` to restore samples — and skip the sample pass instead of aborting, so the run keeps training.

A fuller alternative would be to temporarily reload the real text encoder for the sample pass, but that reintroduces the memory spike this option exists to avoid; graceful skip seemed like the right default. Happy to adjust if you prefer a different behavior.

Related: #1021 (the `cache_text_encoder` device preset missing from `BaseModel`, found on the same run).

### DIFF
--- a/jobs/process/BaseSDTrainProcess.py
+++ b/jobs/process/BaseSDTrainProcess.py
@@ -279,8 +279,31 @@ class BaseSDTrainProcess(BaseTrainProcess):
         # override in subclass
         return generate_image_config_list
 
+    def _text_encoder_is_faked(self) -> bool:
+        from toolkit.unloader import FakeTextEncoder
+
+        text_encoder = getattr(self.sd, "text_encoder", None)
+        encoders = (
+            text_encoder if isinstance(text_encoder, list) else [text_encoder]
+        )
+        return any(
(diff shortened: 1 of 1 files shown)


jaretburkett:
Fixed with 543a0d7

PR #1045, opened 2026-09-16
Add native support for cosine_with_warmup LR scheduler

wcmille:
### Summary
This PR adds native support for the `cosine_with_warmup` (and `cosine_with_hard_restarts_with_warmup`) learning rate scheduler in `toolkit/scheduler.py`, allowing users to configure:
```yaml
train:
  lr_scheduler: "cosine_with_warmup"
  lr_scheduler_params:
    num_warmup_steps: 200  # defaults to 1000 if omitted
    num_cycles: 0.5        # optional, defaults to 0.5
```

---

### Justification & Issue with Prior Implementation

Prior to this change, `get_cosine_schedule_with_warmup` from `diffusers.optimization` was completely unreachable in `ai-toolkit`:

1. **Explicit `"cosine"` branch intercepted Diffusers' lookup**:
   In Diffusers' `SchedulerType` mapping, `SchedulerType.COSINE` maps to `get_cosine_schedule_with_warmup`. However, `toolkit/scheduler.py` explicitly intercepted `"cosine"`:
   ```python
   if name == "cosine":
       if 'total_iters' in kwargs:
           kwargs['T_max'] = kwargs.pop('total_iters')
       return torch.optim.lr_scheduler.CosineAnnealingLR(optimizer, **kwargs)
   ```
   This meant configuring `lr_scheduler: "cosine"` always resolved to PyTorch's native `CosineAnnealingLR` (which lacks warmup functionality).

2. **The generic fallback could never reach it**:
   The fallback block at the end of `get_lr_scheduler`:
   ```python
   else:
       name = SchedulerType(name)
       schedule_func = TYPE_TO_SCHEDULER_FUNCTION[name]
       return schedule_func(optimizer, **kwargs)
   ```
   was unreachable for cosine warmup because:
   - Passing `name: "cosine"` was already caught by the `if name == "cosine"` check above.
   - Passing `name: "cosine_with_warmup"` failed at `SchedulerType(name)` because `"cosine_with_warmup"` is not a valid enum value in Diffusers' `SchedulerType` (Diffusers only defines `"cosine"`).
   - Even if `SchedulerType` had accepted it, Diffusers' `get_cosine_schedule_with_warmup` expects `num_training_steps` and `num_warmup_steps`, whereas `ai-toolkit`'s training loop passes `total_iters`, which would lead to a `TypeError` due to mismatched keyword arguments.

### Solution
- Added an explicit handler for `cosine_with_warmup` (and alias `cosine_warmup`) in `toolkit/scheduler.py` that maps `total_iters`/`num_training_steps` and `num_warmup_steps` cleanly into Diffusers' `get_cosine_schedule_with_warmup`.
- Added support for `cosine_with_hard_restarts_with_warmup`.
- Handled `total_iters` popping for `step` and `constant_with_warmup` to prevent argument collisions.
- Added comprehensive unit tests in `testing/test_scheduler.py`.


wcmille:
Alternately, we could just augment cosine and cosine_with_restarts with a num_warmup_steps that defaults to 0?

wcmille:
Let me take another stab at this where I don't create new lr_schedulers. I think it will be better to simply extend cosine rather than making a cosine_with_warmup.

PR #1049, opened 2026-09-17
Update YuE2 tokenizer_head_joint and nar_lora_joint

hetima:
## Description

The community model repo `Mothersuperior/yue2-mothersuperior-realaudio-tokenizer-v4`
has received new releases, so this updates the YuE2 extension to the current
latest joint pair (v9). The new files also switched from `.pt` to
`.safetensors`, and the safetensors variants use a different internal layout,
so this PR covers the format change as well.

## Changes

- Point `HEAD_FILE` / `NAR_LORA_FILE` at the v9 pair
  (`tokenizer_head_joint_v9.safetensors`, `nar_lora_joint_v9.safetensors`).
  The HF repo name still says "v4" — it hosts every version, and v9 is the
  latest; its metadata marks the pair explicitly
  (`pair_with: tokenizer_head_joint_v9`).
- `SemanticTokenizer` head loading: `.safetensors` files load via
  `safetensors.torch.load_file` (the safetensors repack is a bare state
  dict). The `.pt` path is kept as a fallback.
- `merge_nar_lora`: the safetensors NAR adapter ships flat named keys
  (`layers.N.nar_self_attn/nar_mlp.<proj>.lora_A/B`, plus complete
  `vae2llm`/`llm2vae` replacement weights) instead of the positional tensor
  list + nested io dict the fold loop consumed. The loader rebuilds the
  positional order, so the fold loop itself is unchanged. The `.pt` loader
  is kept as-is.

## Known limitation

The head/adapter file names are still hardcoded constants, so a future
release from that repo will need another small update here. Only the plain
fp32 layout is handled — the `_comfyui` and `.bf16` variants of the same
files are not supported.


jaretburkett:
Fixed in #1055

Issue #1052, opened 2026-09-20
YuE2 trainer: notes from four runs on Windows (step-time collapse fix, CUDA-graph sampling crash) + a question about high-frequency "burn" on small sets


vdruts:
Thanks for the YuE2 support. The joint planner + decoder LoRA gives clearly better voice identity than what we had before, and we have published several LoRAs trained with it. Below is what we ran into over four runs, in case it is useful, plus one question.

**Setup:** Windows 11, RTX 5090 32 GB, ai-toolkit 0.13.16 (Pinokio install), `arch: yue2` on the Comfy-Org int8 ConvRot checkpoint, rank 32, lr 5e-5, `ar_lr_multiplier` 0.6, `ar_kl_weight` 0.2, `cot: full`, `abc_dropout` 0.5, `ar_max_tokens` 4000, `train_window_frames` 1500, batch 1, adamw8bit, sampling disabled. Datasets of 15 to 34 songs.

### 1. Step time collapses from ~4 s to 100+ s on Windows (fix found)

Steps start at 2 to 5 s, then degrade to 100+ s within a few hundred steps while "Shared GPU memory" grows to ~20 GB. It looks like caching-allocator fragmentation from the variable song lengths. On Windows the driver's system-memory fallback means there is never an OOM, so the cache never gets flushed and everything pages over PCIe. `PYTORCH_ALLOC_CONF=garbage_collection_threshold` did not help.

What fixed it: a `torch.cuda.empty_cache()` right after `self._note_train_call()` in `get_noise_prediction` (`extensions_built_in/audio_models/yue2/yue2_model.py`). VRAM went from pinned at 32 GB to about 18 GB and steps stayed at 3 to 7 s for whole runs. Happy to send a PR if you want it (maybe behind a model_kwarg).

### 2. In-training sampling with CUDA graphs crashes the next train step

With sampling enabled and the default AR CUDA graphs, the first train step after a sample fails with `Offset increment outside graph capture`. Workaround: `ar_cuda_graphs: false` in `model_kwargs` (eager samples then take about 6 minutes each), or disable sampling.

### 3. Minor: first start dies after the MERT download

The first launch can die with the hf_hub error "client has been closed" right after the long MERT download. Restarting the job works.

### 4. Question: high-frequency "burn" that grows with passes per song

[...]

PR #1055, opened 2026-09-21
YuE2: support the safetensors-only v5/v8/v9 semantic heads, key latent cache by head

vdruts:
The YuE2 trainer tokenizes training audio with the community semantic head from `Mothersuperior/yue2-mothersuperior-realaudio-tokenizer-v4`. That repo now also carries the v5 / v8 / v9 joint pairs (same architecture; v8/v9 were co-trained with an audio-domain loss against the original recordings, and upstream prefers v9 by ear). They are published as **safetensors only**, and the current loaders only read the original `.pt` layout, so `model_kwargs.semantic_head_path` cannot point at them today.

This keeps v4 as the default and makes the newer files usable:

- **`src/tokenizer.py`**: `load_head_state_dict()` reads either the `.pt` (`{"model": ...}`) or a `.safetensors` head (the flat state dict).
- **`src/model.py`**: `merge_nar_lora` also accepts the named-tensor `.safetensors` NAR adapter and rebuilds the layer-major A/B list + `io` weights the merge already walks. The `_comfyui` fused layout is rejected with a clear error.
- **`yue2_model.py`**: the latent cache stores the head's codec tokens, so a non-default `semantic_head_path` now gets its own latent-space version (`..._tokenizer-head-joint-v9`). The default path keeps the existing key, so nobody's current cache is invalidated. Without this, switching heads silently reuses the old head's tokens.

Usage:

```yaml
model_kwargs:
  semantic_head_path: Mothersuperior/yue2-mothersuperior-realaudio-tokenizer-v4/tokenizer_head_joint_v9.safetensors
  # only if merge_nar_lora is on:
  nar_lora_path: Mothersuperior/yue2-mothersuperior-realaudio-tokenizer-v4/nar_lora_joint_v9.safetensors
```

### Tested (Windows, RTX 5090, torch 2.x)

- v4 head: `.safetensors` vs `.pt` state dicts are bit-exact through the new loader; v4 and v9 safetensors both `load_state_dict(strict=True)` into `TokenizerHead`.
- v4 NAR adapter: the list rebuilt from `.safetensors` is bit-exact with the `.pt` (`lora` 392 tensors + `vae2llm` / `llm2vae`); the v9 adapter loads with the same shapes; the `_comfyui` file raises.
- End to end through `SemanticTokenizer.tokenize` on three full songs: v4 and v9 agree on only 22-31% of frames, so the head choice really does change the training targets, which is why the cache key matters.
- Cache key: default -> unchanged, default + separation -> `_sep` as before, v9 -> `_tokenizer-head-joint-v9`, v9 + separation -> `_tokenizer-head-joint-v9_sep`.

Not tested: a full training run on v9 tokens, and the `merge_nar_lora` path on the bf16 checkpoint beyond the tensor-level equivalence above. I have not changed the default because I have no listening A/B yet showing v9 tokens train a better LoRA; happy to flip it (or add a UI dropdown) if you want that in this PR.

Related: #1052


vdruts:
Follow-up: ran this branch through a real job to close the "not tested" gap on the training path.

- 15-song dataset that already had a v4 latent cache, same config as an earlier run, only `model_kwargs.semantic_head_path` pointed at a local `tokenizer_head_joint_v9.safetensors`.
- The cacher built 15 new files with `latent_space_version: yue2_tokenizer-head-joint-v9` in their metadata and left the 15 existing `yue2` files untouched, so both caches now coexist in `_latent_cache`. Spot check on one song: the old and new `dto.tokens` agree on about 25 % of frames.
- Training then ran the full 350 steps (int8 convrot, `cot: full`, ~3 s/it on a 5090) and saved every checkpoint normally.

Still not exercised: `merge_nar_lora` with the safetensors adapter on the bf16 checkpoint (only the tensor-level equivalence with the `.pt` is verified). Listening comparison of v4-head vs v9-head LoRAs is rendering now; I'll report if it says anything useful about which head should be the default.

